PR #220, opened 2024-11-04
DepID: use § to encode /

isaacs:
This fixes an issue where 11ty's module loader gets upset at seeing %2F in a module path, thinking that it must be a sneaky way to escape their security sandbox.

This is backwards compatible, but it *will* leave garbage behind in node_modules/.vlt if upgraded to without cleaning out node_modules.

It would be good, as a future enhancement, to have a job that kicks off at reification to remove any unlinked trash that might be living in the .vlt folder. This would also defend against unfinished folder removals which currently can potentially leave behind a bunch of `.VLT.DELETE.*`


isaacs:
Force-pushed to make the root-level `snap` and `test` targets work properly, and to update the browser form of depid as well as the server form of it.

PR #239, opened 2024-11-15
registry-client: retry on transient network fails

isaacs:
This fixes an issue where we'd crash when the network is flaky, with `ECONNREFUSED`, `ECONNRESET`, `EPIPE`, and `UND_ERR_SOCKET`.

Using undici's `RetryHandler` takes care of this transparently, including retrying `GET` requests with a `Range` header (for servers that advertise support for it) if the body had already been partly consumed.

### DIFF
--- a/src/registry-client/src/index.ts
+++ b/src/registry-client/src/index.ts
@@ -3,7 +3,7 @@ import { register } from '@vltpkg/cache-unzip'
 import { type Integrity } from '@vltpkg/types'
 import { XDG } from '@vltpkg/xdg'
 import { loadPackageJson } from 'package-json-from-dist'
-import { Agent, type Dispatcher } from 'undici'
+import { Agent, RetryAgent, type Dispatcher } from 'undici'
 import { addHeader } from './add-header.js'
 import { CacheEntry } from './cache-entry.js'
 import { bun, deno, node } from './env.js'
@@ -19,6 +19,17 @@ export type RegistryClientOptions = {
    * @default `$HOME/.config/vlt/cache`
    */
   cache?: string
+  /**
+   * Number of retries to perform when encountering network errors or
+   * likely-transient errors from git hosts.
+   */
+  'fetch-retries'?: number
+  /** The exponential backoff factor to use when retrying git hosts */
+  'fetch-retry-factor'?: number
+  /** Number of milliseconds before starting first retry */
+  'fetch-retry-mintimeout'?: number
+  /** Maximum number of milliseconds between two retries */
+  'fetch-retry-maxtimeout'?: number
 }
 
 export type RegistryClientRequestOptions = Omit<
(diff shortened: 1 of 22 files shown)


src/registry-client/src/index.ts
```diff
@@ -91,19 +102,41 @@ const agentOptions: Agent.Options = {
...
-    cache = xdg.cache('registry-client'),
-  }: RegistryClientOptions) {
+  constructor(options: RegistryClientOptions) {
+    const {
+      cache = xdg.cache('registry-client'),
+      'fetch-retry-factor': timeoutFactor = 2,
+      'fetch-retry-mintimeout': minTimeout = 0,
+      'fetch-retry-maxtimeout': maxTimeout = 30_000,
+      'fetch-retries': maxRetries = 3,
+    } = options
```
lukekarrys, line 115: Generally, I'm not a fan of doubling up on these defaults in both the CLI and the workspace. It's already a pattern we use everywhere so I'm ok with adding another instance here.

I'd rather have our workspaces be less ergonomic and rely on types to force setting the options. I think this pattern is less likely to lead to bugs in the future if CLI defaults get changed or removed.

Again, not blocking this PR but curious to pros to this pattern that I might not be seeing.

lukekarrys (approved):

src/registry-client/src/index.ts
isaacs: Good call. Updated, PTAL!

src/vlt/src/config/index.ts
```diff
@@ -321,7 +317,7 @@ export class Config {
     if (this.command) p.positionals.shift()
     else this.command = getCommand(p.values['fallback-command'])
 
-    return Object.assign(this, p)
+    return Object.assign(this, p) as this & ParsedConfig
```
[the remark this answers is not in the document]
lukekarrys: I thought about adding a code like `ETHISSHOULDNEVERHAPPEN` where the error message would be "we messed up please file a bug".

lukekarrys (approved):

src/vlt/src/config/index.ts
```diff
@@ -321,7 +317,13 @@ export class Config {
     if (this.command) p.positionals.shift()
     else this.command = getCommand(p.values['fallback-command'])
 
-    return Object.assign(this, p)
+    Object.assign(this, p)
+
+    /* c8 ignore start - unpossible */
+    if (!isParsed(this)) throw error('failed to parse config')
+    /* c8 ignore stop */
```
lukekarrys, line 324:
```suggestion
    /* c8 ignore next - unpossible */
    if (!isParsed(this)) throw error('failed to parse config')
```

src/vlt/src/config/index.ts
```diff
@@ -321,7 +317,13 @@ export class Config {
...
+
+    return this as this & ParsedConfig
```
lukekarrys, line 326:
```suggestion
    return this
```

PR #241, opened 2024-11-17
graph: fix adding nameless specs 

ruyadorno:
This PR fixes adding nameless dependency specs when using the cli
e.g: `vlt install github:foo/bar` along with fixing inconsistencies
when merging cli-defined dependencies with manifest-defined values.

A few connected changes found when turning many stones related to this
problem that are worthy of mention:

- `Graph.removeNode` removes the edge from its `from` node instead of
  just unsetting the `to` value of that edge.
- `append-nodes` will fix up and update entries in the user-provided
  `add` values, so that we can add a proper dependency name when writing
  to that importer `package.json` at the end of reify.
- cleaned up the logic between `ideal.build` and its
  `get-importer-specs` module so values should be properly merged when
  adding / removing new dependencies.
- Make lockfile load a little more resilient by making sure it does not
  throw when trying to deconstruct values from a missing `options`
  property.
- updated a few tests and snapshots that had either wrong assumptions
  and / or bad fixture data

### DIFF
--- a/src/graph/src/ideal/append-nodes.ts
+++ b/src/graph/src/ideal/append-nodes.ts
@@ -4,6 +4,7 @@ import { type PackageInfoClient } from '@vltpkg/package-info'
 import { Spec, type SpecOptions } from '@vltpkg/spec'
 import { type PathScurry } from 'path-scurry'
 import {
+  asDependency,
   longDependencyTypes,
   shorten,
   type Dependency,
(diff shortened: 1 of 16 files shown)


src/graph/src/ideal/append-nodes.ts
```diff
@@ -99,6 +101,26 @@ export const appendNodes = async (
           throw er
         })
 
+      // when an user is adding a nameless dependency, e.g: `github:foo/bar`,
+      // `file:./foo/bar`, we need to update the `add` option value to set the
+      // correct name once we have it, so that it can properly be stored in
+      // the `package.json` file at the end of reify.
+      if (mani?.name && spec.name.startsWith('(unknown)')) {
```
isaacs, line 108: Why not just this?

```suggestion
      if (mani?.name && spec.name === '(unknown)') {
```

ruyadorno: done ✅

PR #250, opened 2024-11-27
chore(.nvmrc): add node version 20

sukvvon:
* specify node version with `.nvmrc`
* update `actions/setup-node` to use `node-version-file: '.nvmrc'` instead `node-version`


.nvmrc
```diff
@@ -0,0 +1 @@
+v20.18.1
```
darcyclarke, line 1:
```suggestion
20

```

lets change this back to a greedy version range (ie. `20.x`) but since nvm doesn't support `x` we can just set this to `"20"` & it should do the same thing.

darcyclarke (changes requested):

Issue #176, opened 2024-10-16
bug: root detection in a nested directory inside git repo


lukekarrys:
```
❯ tree -a -L2
.
├── .git
│   ├── HEAD
│   ├── config
│   ├── description
│   ├── hooks
│   ├── info
│   ├── objects
│   └── refs
└── nested
    └── package.json

7 directories, 4 files
❯ cd nested
❯ cat package.json
   1   │ {
   2   │   "dependencies": {
   3   │     "abbrev": "*"
   4   │   }
   5   │ }
❯ vlt i
file:///Users/lukekarrys/.npm-bins/node_modules/vlt/commands/install.js:38227
  const mainManifest = conf.options.packageJson.read(conf.options.projectRoot);
                                                ^

Error: Could not read package.json file
    at command (file:///Users/lukekarrys/.npm-bins/node_modules/vlt/commands/install.js:38227:49)
    at Object.run (file:///Users/lukekarrys/.npm-bins/node_modules/vlt/vlt.js:12175:15) {
  [cause]: {
    path: '/Users/lukekarrys/Desktop/scratch/vlt-root-detection/package.json',
    cause: Error: ENOENT: no such file or directory, open '/Users/lukekarrys/Desktop/scratch/vlt-root-detection/package.json'
        at readFileSync (node:fs:442:20)
        at PackageJson.read (file:///Users/lukekarrys/.npm-bins/node_modules/vlt/vlt.js:219:40)
        at command (file:///Users/lukekarrys/.npm-bins/node_modules/vlt/commands/install.js:38227:49)
        at Object.run (file:///Users/lukekarrys/.npm-bins/node_modules/vlt/vlt.js:12175:15) {
      errno: -2,
      code: 'ENOENT',
      syscall: 'open',
      path: '/Users/lukekarrys/Desktop/scratch/vlt-root-detection/package.json'
    }
  }
}

Node.js v22.4.0
```

lukekarrys:
I don't think this is actually a bug. I had forgotten we designed root detection to use a `.git` directory as well. There is a better error message that we can show here though. I think we should make it clear:

- what directory what root was detected
- why that root was detected
- a more clear message when the root has no package.json file

darcyclarke:
Had no clue this existed already but https://github.com/vltpkg/vltpkg/pull/251 should address this now

PR #251, opened 2024-12-03
fix: project root resolution

darcyclarke:
### Description
- fixes root resolution which was causing errors to occur when trying to run `vlt` inside a nested dir of any git project
- the root-cause was incorrectly treating the existence of `.git` as both a backstop & a root identifier
- this change also... 
   - resolves issues where a previously found/known-good root existed while walking up the path but was ignored
   - treats the cwd as the preferred default root instead of the `.git` backstop (which is less confusing for end-users if no known root is found while walking up the path)
- the existence of `vlt-workspaces.json` continues to be treated as both a backstop & a well-known root location

### Minimum Reproduction
- since I'm not opening a separate issue for the resolution error/backstop, here's a minimum reproduction to see the issue with the previous algorithm

```bash
mkdir test && cd test # make a test dir
git init # initial git repo
mkdir foo && cd foo # make a subdir
npm init -y # create a package.json
vlt install express # try to install & get errors as vlt uses an incorrect root location
```

Screenshot 2024-12-03 at 11 35 08 AM

### DIFF
--- a/src/vlt/src/config/index.ts
+++ b/src/vlt/src/config/index.ts
@@ -582,25 +582,29 @@ export class Config {
     // indicators that this *may* be the root, if no .git or workspaces
     // file is found higher up in the search.
     let lastKnownRoot = this.projectRoot
-    for (const dir of walkUp(this.projectRoot)) {      
+    for (const dir of walkUp(this.projectRoot)) {
       // don't look in ~
       if (dir === home) break
 
       // finding a project config file stops the search
       const projectConfig = resolve(dir, 'vlt.json')
       if (projectConfig === userConfig) break
-      if (await exists(projectConfig) && await this.#maybeLoadConfigFile(projectConfig)) {
+      if (
+        (await exists(projectConfig)) &&
+        (await this.#maybeLoadConfigFile(projectConfig))
+      ) {
         lastKnownRoot = dir
         break
       }
 
       // stat existence of these files
-      const [hasPackage, hasModules, hasWorkspaces, hasGit] = await Promise.all([
-        exists(resolve(dir, 'package.json')),
-        exists(resolve(dir, 'node_modules')),
-        exists(resolve(dir, 'vlt-workspaces.json')),
-        exists(resolve(dir, '.git'))
-      ])
+      const [hasPackage, hasModules, hasWorkspaces, hasGit] =
+        await Promise.all([
+          exists(resolve(dir, 'package.json')),
+          exists(resolve(dir, 'node_modules')),
+          exists(resolve(dir, 'vlt-workspaces.json')),
+          exists(resolve(dir, '.git')),
+        ])
 
       // treat these as potential roots
       if (hasPackage || hasModules || hasWorkspaces) {
--- a/src/vlt/src/config/index.ts
+++ b/src/vlt/src/config/index.ts
@@ -581,7 +581,7 @@ export class Config {
 
     // indicators that this *may* be the root, if no .git or workspaces
     // file is found higher up in the search.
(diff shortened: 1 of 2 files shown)


src/vlt/src/config/index.ts
```diff
@@ -579,41 +579,44 @@ export class Config {
     const userConfig = xdg.config('vlt.json')
     await this.#maybeLoadConfigFile(userConfig)
 
-    // don't walk up past a folder containing any of these
-    const stops = ['vlt-workspaces.json', '.git']
     // indicators that this *may* be the root, if no .git or workspaces
```
isaacs, line 582: This comment should go too, yes? Or does it apply to `lastKnownRoot`?

isaacs (approved):
This looks good to me.

Just to confirm, you _can't_ have a `vlt.json` in a workspace, correct?

I like this restriction, personally. But that said (and this is 100% a "future patch" suggestion), we may need to figure out a way to have another config layer (i knowwwww ughhhh) at the _individual workspace_ level, that overrides the configs in the `projectRoot`. For example, I could easily imagine a situation where some workspaces in a project need to publish to an internal registry, and some to the public registry. If/when that's needed, we can mitigate the complexity (or maybe more like "push it onto the user", sadly) by making the config be in a different file (or even in a field in `package.json`, since it's single-package/workspace-specific).

But regardless, this PR cleans up the logic here nicely, and sets it up to make that future change more easily, so it's a good step in the right direction 👍

isaacs:
Marked as "approved", but realized the comment about the misleading comment should probably be fixed. But other than that, LGTM.

src/vlt/src/config/index.ts
darcyclarke, line 582:
```suggestion
```

src/vlt/src/config/index.ts
```diff
@@ -579,41 +579,43 @@ export class Config {
     const userConfig = xdg.config('vlt.json')
     await this.#maybeLoadConfigFile(userConfig)
 
-    // don't walk up past a folder containing any of these
-    const stops = ['vlt-workspaces.json', '.git']
-    // indicators that this *may* be the root, if no .git or workspaces
     // file is found higher up in the search.
```
darcyclarke, line 582:
```suggestion
```

src/vlt/src/config/index.ts
darcyclarke: Ya, makes sense to remove it

isaacs:
LGTM!

lukekarrys:
I believe this closes https://github.com/vltpkg/vltpkg/issues/176

src/vlt/src/config/index.ts
```diff
@@ -579,41 +579,42 @@ export class Config {
...
-        this.projectRoot = dir
+
+      // stat existence of these files
+      const [hasPackage, hasModules, hasWorkspaces, hasGit] =
+        await Promise.all([
+          exists(resolve(dir, 'package.json')),
+          exists(resolve(dir, 'node_modules')),
+          exists(resolve(dir, 'vlt-workspaces.json')),
+          exists(resolve(dir, '.git')),
+        ])
```
lukekarrys, line 605: I might do a follow up to see if a `Promise.race` here has a measurable perf improvement.

lukekarrys (approved):

src/vlt/src/config/index.ts
lukekarrys: Ignore me, this wouldn't do anything because we aren't stopping at this point. We'd still need to make all the `exists` calls eventually.

PR #268, opened 2024-12-12
chore(package.json): add packageManager pnpm@9.12.0

sukvvon:
* add `packageManager: pnpm@9.12.0` in `package.json`
* in `pnpm/action-setup`, if `packageManager` is existed, we dont need to set `version`. `pnpm/action-setup` follow package.json's `packageManager`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -31,7 +31,6 @@ jobs:
 
       - uses: pnpm/action-setup@v4
         with:
-          version: 9
           run_install: false
 
       - name: Setup Nodejs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,6 @@ jobs:
 
       - uses: pnpm/action-setup@v4
         with:
-          version: 9
           run_install: false
 
       - name: Setup Nodejs
@@ -145,7 +144,6 @@ jobs:
 
       - uses: pnpm/action-setup@v4
         with:
-          version: 9
           run_install: false
 
       - name: Use Nodejs ${{ matrix.node-version }}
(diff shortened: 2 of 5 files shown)


darcyclarke:
`packageManager` doesn't support ranges &, again, it's cumbersome to maintain explicit/pinned versions of a dep we'd want to ensure gets timely patches for (ex. security/bug fixes). I think we're fine to keep `version` for now. Appreciate the help trying to optimize these things for us though. Going to close this unless you think there's a better alternative.

sukvvon:
@darcyclarke 

the reason for explicitly adding the packageManager field in package.json is as follows:

### preventing unnecessary package.json modifications

without specifying packageManager, running pnpm install automatically adds the currently used version of pnpm to package.json. this results in unintended git diff changes, creating unnecessary confusion during code reviews and version management. to prevent this, the packageManager field is explicitly set to define the package manager and version used by the project.

### maintaining environmental consistency

it is essential to ensure that all team members and ci/cd environments use the same package manager and version. without explicitly specifying this, the behavior of pnpm install can vary depending on the local environment and version of pnpm installed by each developer. this can lead to unintended dependency conflicts or build failures.
 
### managing version upgrades

while there may be a desire to always use the latest pnpm version, new versions can introduce unexpected issues. by specifying the packageManager field, the team can enforce the use of a stable, well-tested version and manually upgrade as needed, ensuring potential risks are managed proactively.

### ensuring ci/cd pipeline stability

without specifying packageManager, ci environments may default to using the latest version of pnpm. this can cause inconsistencies in behavior due to version differences, potentially leading to build failures. explicitly specifying the version prevents such issues at their root.

darcyclarke:
I'm very aware of what it does, thank you. We're not going to add it at this time

PR #277, opened 2024-12-14
Update dev deps and lint fixes

lukekarrys:
This PR updates all of our dev deps (except typescript because the new version isn't compatible with `tap` I think) and a few other commits that fix new lint errors caught by the upgrade to `typescript-eslint`.

Changes:
- Use a new script for updating deps based on `npm-check-updates`
- Update promise-spawn types to get better narrowing on results
- Fix new promise-spawn types' typedoc generation for some unresolved links
- Ignore some new errors of [`no-base-to-string`](https://typescript-eslint.io/rules/no-base-to-string/) because they are probably fine


src/query/src/attribute.ts
```diff
@@ -5,6 +5,11 @@ import { asAttributeNode, type ParserState } from './types.js'
 
 export type ComparatorFn = (attr: string, value?: string) => boolean
 
+// JSONField has a mapped type constituent that would coerce to [object Object]
+// when stringified, which is what we want in this case.
```
lukekarrys, line 9: @ruyadorno Is this comment I added accurate? There is a test case that asserts we convert an object to the string `"[object Object]"`, so I wanted to check that this is not a bug.

ruyadorno: yes 👍 that LGTM, the idea is that we want a direct match to the attribute provided in the user query, so an object being converted to `"[object Object]"` since it won't match user input

scripts/update-deps.js
```diff
@@ -1,190 +1,145 @@
...
+      ...Object.entries(values).map(([k, v]) => `--${k}=${v}`),
+    ].join(' '),
+    { encoding: 'utf-8' },
+  )
+
+  let jsonOutput = null
+  try {
+    jsonOutput = JSON.parse(res)
+  } catch {
+    console.log(res)
```
ruyadorno, line 105: do we even want to `console.log` something in here? if the answer is yes, than may I suggest using `console.warn('<here is the reason>:', res)` instead?

lukekarrys: That was attempting to still allow for `--help` to be passed to `npm-check-updates` which will always return a string and not json.

But I reworked it to use the programmatic API and only fallback to getting the help output if `--help` is passed explicitly.

PR #292, opened 2024-12-16
deps: typescript@5.7.2

lukekarrys:


lukekarrys:
Our tap tests are no longer passing after updating TypeScript. This may require other work in tap to unblock updating. Closing since we are ok on the current version of TypeScript and I don't plan to get this passing in its current form.

PR #307, opened 2024-12-21
error-cause: allow cause to be unknown

lukekarrys:
We turned on `useUnknownInCatchVariables` in our tsconfig when we started linting 9cc25431b5567aa5c094c4557b7626df75a74d41. So I think it makes sense for the `cause` property on `ErrorCauseObject` to be `unknown` as well. This removes the need down to do `{ cause: er as Error }` every time we want to throw a previously caught error.

I originally set the `cause: ErrorCause` argument for all the exported `error|typeError|syntaxError()` functions to also be `unknown` but this gets rid of some of the type safety. So you still have to pass an actual `Error | ErrorCauseObject` to the functions, but the cause object can now have `{ cause: unknown }`.

This also adds some type guard utility functions to `error-cause` which I used to clean up some `er instanceof Error` checks.

### DIFF
--- a/src/registry-client/src/index.ts
+++ b/src/registry-client/src/index.ts
@@ -216,19 +216,11 @@ export class RegistryClient {
 
       if (isRedirect(result)) {
         resp.body.resume()
-        try {
-          const [nextURL, nextOptions] = redirect(options, result, u)
-          if (nextOptions && nextURL) {
-            return this.request(nextURL, nextOptions)
-          }
-          return result
-        } catch (er) {
-          /* c8 ignore start */
-          throw er instanceof Error ? er : (
-              new Error(typeof er === 'string' ? er : 'Unknown error')
-            )
-          /* c8 ignore stop */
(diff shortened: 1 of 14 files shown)


src/registry-client/src/index.ts
```diff
@@ -216,19 +216,11 @@ export class RegistryClient {
...
-          if (nextOptions && nextURL) {
-            return this.request(nextURL, nextOptions)
-          }
-          return result
-        } catch (er) {
-          /* c8 ignore start */
-          throw er instanceof Error ? er : (
-              new Error(typeof er === 'string' ? er : 'Unknown error')
-            )
-          /* c8 ignore stop */
```
lukekarrys, line 230: This removes the catch block entirely which is not directly connected to the rest of the error-cause changes but I did notice this while grepping for `instanceof Error`.

This catch block is just re-throwing the same error unless it's a weird error in which case it tries to normalize it. We don't need to normalize it here and can instead catch and normalize it once we truly handle it.

ruyadorno (approved):

Issue #266, opened 2024-12-12
[BUG] pinned dep changed to use `^` on `vlt i`


ljharb:
my packages have `"eslint": "=8.8.0"`, in dev deps. the pinning is required for peer dep requirements. (the reason it's a peer dep requirement isn't relevant). when i `vlt i`, it changes the `=` to a `^` in package.json, which is very very bad.

In other words, the prefix (`''`, `=`, `^`, `>=`, etc) should probably never be changed without the user explicitly requesting it.

darcyclarke:
Should be fixed now that #310 & #311 have landed - we'll get a new release out imminently.

PR #311, opened 2025-01-09
Only updates `package.json` when there are modified dependencies

ruyadorno:
Adds a `modifiedDependencies` companion flag to the `AddImportersDependenciesMap` and `RemoveImportersDependenciesMap` objects that tracks if there are any dependencies being add / removed in any of the importers.

The initial usage of this flag is to skip writing to `package.json` at the end of a reify in case there were no deps add or removed.

Relates to: https://github.com/vltpkg/vltpkg/issues/266


src/graph/src/ideal/get-importer-specs.ts
```diff
@@ -114,6 +131,14 @@ export const getImporterSpecs = ({
     graph,
   })
 
+  // set the modifiedDependencies flag if any
+  // of the importers have modified dependencies
+  for (const addDeps of addResult.values()) {
+    if (addDeps.size > 0) {
+      addResult.modifiedDependencies = true
```
lukekarrys, line 138: we could `break` after this right?

lukekarrys (dismissed):

PR #316, opened 2025-01-12
registry-client: remove v8 serialization

ruyadorno:
Removes the special v8 serialization in favor of using only standard JSON parse / stringify.

On benchmarks the v8 serialization of json files is not showing any benefits and this extra info comes at a cost of twice the cache size for all cached json files and the extra maintainance / complexity.

### Benchmarks

```
hyperfine --warmup 2 --prepare="rm -rf node_modules" --setup="rm -rf /Users/ruyadorno/Library/Caches/vlt" -r=16 --cleanup="sleep 1" "vlt install" "vlt-main install"
Benchmark 1: vlt install
  Time (mean ± σ):      2.292 s ±  0.080 s    [User: 3.309 s, System: 2.320 s]
  Range (min … max):    2.150 s …  2.421 s    16 runs

Benchmark 2: vlt-main install
  Time (mean ± σ):      2.302 s ±  0.095 s    [User: 3.113 s, System: 2.412 s]
  Range (min … max):    2.198 s …  2.533 s    16 runs

Summary
  vlt install ran
    1.00 ± 0.05 times faster than vlt-main install
```


src/registry-client/src/cache-entry.ts
```diff
@@ -267,17 +264,8 @@ export class CacheEntry {
...
-      /* c8 ignore start - very rare, but theoretically possible to throw */
-      try {
-        return (this.#json = deserialize(ser) as JSONObj)
-      } catch {}
-      /* c8 ignore stop */
-    }
     const obj = JSON.parse(this.text()) as JSONObj
-    if (serializedHeader)
-      this.setHeader(serializedHeader, serialize(obj))
+    this.#json = obj
```
ruyadorno, line 268: This line is already handled in https://github.com/vltpkg/vltpkg/pull/314 but it's vital for the sake of running these benchmarks since non-v8-serialized-json objects were simply never getting memoized.

isaacs (dismissed):

Issue #271, opened 2024-12-12
[FEATURE] GUI: Dashboard should find all projects on a system


darcyclarke:
### Description
Currently, the algorithm to find projects only does a minimal walk of the cwd/path to find relevant projects but this is confusing & will be hard to communicate to end-users why/what gets displayed with this current algo.

### Suggestion
- [ ] change the lookup algorithm to find **all** "projects" on a given system
- [ ] cache previously seen/known project locations for faster initial lookups & then update/load the state in the background for all other locations
- [ ] test rendering a loading state & waiting for up-to-date state vs. loading previous known state & loading in the background (ie. consider how we communicate we're running some background process to persist the state of the system with the ui experience)

PR #319, opened 2025-01-17
vlt: add new dashboard root path logic

ruyadorno:
This changeset introduces a new system for finding out projects to be displayed in the GUI dashboard.

It's now configurable via a `dashboard-root` config option and otherwise defaults to reading all projects available starting from the user home directory.

Fixes: https://github.com/vltpkg/vltpkg/issues/271

### DIFF
--- a/src/vlt/src/start-gui.ts
+++ b/src/vlt/src/start-gui.ts
@@ -242,11 +242,17 @@ const updateGraphData = (
 
 const updateDashboardData = async (
   tmp: string,
-  options: ConfigOptions,
+  conf: LoadedConfig,
 ) => {
+  const userDefinedProjectPaths =
+    // eslint-disable-next-line
+    conf.values?.['dashboard-root'] || []
   const dashboard = formatDashboardJson(
-    readProjectFolders(process.cwd(), options),
-    options,
+    readProjectFolders({
+      ...conf.options,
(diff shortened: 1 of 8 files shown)


src/vlt/src/start-gui.ts
```diff
@@ -224,11 +224,17 @@ const updateGraphData = (
 
 const updateDashboardData = async (
   tmp: string,
-  options: ConfigOptions,
+  conf: LoadedConfig,
 ) => {
+  const userDefinedProjectPaths =
+    // eslint-disable-next-line
```
ruyadorno, line 230: @lukekarrys had to end up adding this disable line comment otherwise eslint keeps removing the `?.` from the next line and tests will actually break without that 😕

darcyclarke (approved):

src/vlt/src/start-gui.ts
lukekarrys: Looks like we're doing some manual casting with `as LoadedConfig` in the tests. According to the types `LoadedConfig['values']` always exists but when we cast we're suppressing that error in the tests. The autofix behavior is following the types, so it "fixes" it which actually breaks.

I added a band-aid in #335. We currently have a lot of linting rules turned off for tests. I wonder if those would help spot the unsafe behavior in these cases?

PR #337, opened 2025-01-26
fix(gui): text-overflow on dashboard item paths

lukekarrys:
## Before
Screenshot 2025-01-26 at 10 38 52 AM

## After
Screenshot 2025-01-26 at 1 15 39 PM

### DIFF
--- a/src/gui/src/components/dashboard-grid/index.tsx
+++ b/src/gui/src/components/dashboard-grid/index.tsx
@@ -117,19 +117,15 @@ export const DashboardItem = ({
 
       <div className="w-full h-12 flex items-center gap-4 justify-between border-t-[1px] px-4 py-3">
         <TooltipProvider>
-          <div className="flex items-center justify-center rounded-sm border-[1px] px-2 py-1">
-            <Tooltip>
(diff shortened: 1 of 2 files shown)


ruyadorno (changes requested):
Great catch! The only thing I believe needs to be changed is to crop + ellipsis in the middle, since usually the first part and last part of a pathname are usually the most important ones. wdyt?

lukekarrys:
I wanted to find a CSS only solution to address the overflow issues and think any path cropping can be addressed in a future PR.

We'll always need the CSS solution in place, but I do agree we can come up with better heuristics like:
- replacing HOME with `~`
- removing the longest common prefix which is especially helpful when using a custom `--dahboard-root`
- shortening individual path segments like `fish` does

ruyadorno (approved):
> I wanted to find a CSS only solution to address the overflow issues and think any path cropping can be addressed in a future PR.

Sure 👍 SGTM then!

ruyadorno:
ok, opened https://github.com/vltpkg/vltpkg/pull/342 to follow up with homedir shortener then! 👍

PR #338, opened 2025-01-26
sort tailwind classes with prettier

lukekarrys:
This should not get merged until there is no inflight gui or docs PRs.


ruyadorno:
Yes, let's please wait until there are less active development in the GUI before landing this as I know this week is probably going to be busy with GUI improvements too. 🙏

PR #340, opened 2025-01-27
install: show output in the terminal

lukekarrys:
https://github.com/user-attachments/assets/9bd4e2de-98eb-426c-aff7-f665388468cd


src/vlt/src/output.ts
```diff
@@ -12,11 +14,42 @@ export const stdout = (...args: unknown[]) => console.log(...args)
...
+  }
+  const defaultView =
+    'defaultView' in views ? views.defaultView : undefined
+  const viewName = conf.values.view ?? defaultView
+
+  const commandResult = await command(conf)
+
+  // /* c8 ignore next */
+  // const view = viewName ? views[viewName] : undefined
+
```
ruyadorno, line 44: oops

ruyadorno (commented):
I think this is looking like a good start! 👍

PR #348, opened 2025-01-28
gui: root button

ejkorol:
Creates a back button for the gui view.

Allows a user to quickly jump back to the root view of the project.

Screenshot 2025-01-28 at 15 02 11


ruyadorno (commented):
I really think it should be named either `HomeButton` or `RootButton` instead, Back button is actually very confusing, since it's not going back in the usual history sense nor the icon represents the usual back button arrow 😅

ruyadorno (approved):

PR #355, opened 2025-01-28
query view directory tooltip

### LINKED ISSUE #350: [CHORE] GUI: Add tooltip when hovering query directory values
### Description
Currently when hovering long directory names in the query list view you do not get a tooltip/help text to see what the full path may be.

Image

ejkorol:
Adds tooltip for the directory/context of the query to view untruncated text

Screenshot 2025-01-28 at 16 17 13

Fixes #350

### DIFF
--- a/src/gui/src/components/queries/saved-item.tsx
+++ b/src/gui/src/components/queries/saved-item.tsx
@@ -181,8 +181,8 @@ const SavedQueryItem = ({
         <div className="flex col-span-2 items-center gap-2 w-full">
           <TooltipProvider>
             <Tooltip>
-              <TooltipTrigger className="w-full text-left">
-                <p className="text-sm truncate">{item.context}</p>
+              <TooltipTrigger className="text-sm truncate w-full text-left">
+                {item.context}
               </TooltipTrigger>
               <TooltipContent>{item.context}</TooltipContent>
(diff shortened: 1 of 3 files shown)


src/gui/src/components/queries/saved-item.tsx
```diff
@@ -172,8 +178,15 @@ const SavedQueryItem = ({
...
           )}
         </div>
-        <div className="flex col-span-2 items-center gap-2">
-          <p className="text-sm truncate">{item.context}</p>
+        <div className="flex col-span-2 items-center gap-2 w-full">
+          <TooltipProvider>
+            <Tooltip>
+              <TooltipTrigger className="w-full text-left">
+                <p className="text-sm truncate">{item.context}</p>
+              </TooltipTrigger>
```
lukekarrys, line 186:
```suggestion
              <TooltipTrigger className="text-sm truncate w-full text-left">
                {item.context}
              </TooltipTrigger>
```

I think we can do this same thing all on the `<TooltipTrigger>`. It's one less element and less error prone since the `truncate` on the `<p>` class doesn't always work depending on how the parent gets rendered. That's what happened in #337.

lukekarrys (approved):

PR #367, opened 2025-01-30
gui dashboard cards

ejkorol:
updates styles of dashboard cards


src/gui/test/components/dashboard-grid/__snapshots__/index.tsx.snap
ruyadorno, line 1: can we mock the icons in the `src/gui/test/components/dashboard-grid/index.tsx` ?

ruyadorno (approved):

PR #369, opened 2025-01-31
gui: fix uninstall dependencies dropdown

ruyadorno:
Fixes another issue with z-indexes, both the uninstall dependencies popup and the toaster should be visible now.


ejkorol (approved):
👏

PR #371, opened 2025-01-31
chore: add repository metadata to workspaces

ruyadorno:
Will make it possible to render the favicon for our own workspaces when navigating the GUI and will also make sure that metadata is available in other places such as npmjs.com

### DIFF
--- a/src/package-info/test/index.ts
+++ b/src/package-info/test/index.ts
@@ -907,7 +907,10 @@ t.test('path git selector', async t => {
     )
   })
 
-  await t.test('extract', async t => {
+  // git-cloning in a resource constrained machine running plenty
+  // of tests in parallel might take a lot more time than usual,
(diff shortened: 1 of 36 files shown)


lukekarrys (approved):

lukekarrys (approved):

src/package-info/test/index.ts
```diff
@@ -907,7 +907,10 @@ t.test('path git selector', async t => {
     )
   })
 
-  await t.test('extract', async t => {
+  // git-cloning in a resource constrained machine running plenty
+  // of tests in parallel might take a lot more time than usual,
+  // so we have a generous timeout here to satisfy windows CI.
+  await t.test('extract', { timeout: 10_000 }, async t => {
```
lukekarrys, line 913: I think we should remove this. It still timed out once more with it, and #381 bumped the timeout globally.

PR #373, opened 2025-02-02
Isaacs/token add rm

isaacs:


src/vlt/src/commands/token.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    case 'rm': {
+      await deleteToken(reg, conf.options.identity)
+      break
+    }
+
+    default: {
+      //eslint-disable-next-line no-console
+      console.error(usage().usage())
```
lukekarrys, line 32: here you can `throw error('Not a good token subcommand!!!', { code: 'EUSAGE' })` and the output will take care of displaying the command usage and setting the exit code

src/vlt/src/read-password.ts
```diff
@@ -0,0 +1,41 @@
...
+    stdin.on('data', (c: Buffer) => {
+      // backspace
+      if (c.length === 1 && c[0] === 0x7f) {
+        input = input.substring(0, input.length - 1)
+        stdout.write('\x1b[1D \x1b[1D')
+        return
+      }
+
+      input += String(c)
+      //eslint-disable-next-line no-control-regex
```
lukekarrys, line 25: i think this is a default eslint rule aimed at people not writing CLI applications. if you think its better, we can turn it off altogether?

src/vlt/tap-snapshots/test/commands/help.ts.test.cjs
```diff
@@ -151,6 +151,18 @@ More documentation available at <https://docs.vlt.sh>
 
   --fetch-retry-maxtimeout=<n>
                        Maximum number of milliseconds between two retries
+  -i<name> --identity=<name>
```
lukekarrys, line 154: i cant imagine this is associated with your changes, but why is there not a newline before this?

src/vlt/test/commands/token.ts
```diff
@@ -0,0 +1,53 @@
...
+  const exitCode = process.exitCode
+  const logs = t.capture(console, 'error').args
+  await command({
+    options: { registry: 'https://registry.vlt.javascript/' },
+    positionals: ['wat'],
+  } as unknown as LoadedConfig)
+  t.equal(process.exitCode, 1)
+  process.exitCode = exitCode
+  t.strictSame(logs(), [[usage().usage()]])
+})
```
lukekarrys, line 53: if the command is changed to throw `EUSAGE` instead of log directly, the `mockCommandOutput` helper can be used here which will run the command through the output helper and return an object of `{ errs, logs }` where `errs` will be everything logged to `stderr`

lukekarrys: or it can just assert that it throws. but i found it helpful doing the other commands to be able to test more integration style and get what would be output to the terminal

lukekarrys (dismissed):
👍🏼 Added some comments as FYI, then can be addressed in the future (or never).

src/vlt/tap-snapshots/test/commands/help.ts.test.cjs
isaacs: That's just jackspeak's behavior. It only adds extra newlines when the descriptions wrap. Otherwise, it starts to feel excessively spaced-out.

src/vlt/src/read-password.ts
isaacs: Personally, I feel like pretty much _all_ of the regex-related eslint rules are excessive and patronizing, but people also seem to think regular expressions are hard, and I think they're one of the most simple and straightforward features of JavaScript, so I try to accept my minority status and go along to get along. But yeah, we can definitely remove this one, definitely fine by me.

src/vlt/test/commands/token.ts
isaacs: Yeah, throwing an EUSAGE and asserting it now.

PR #378, opened 2025-02-05
vlt: add init command

ruyadorno:
Add an initial, minimalistic `vlt init` & `vlt pkg init` implementation.


src/vlt/src/commands/pkg.ts
```diff
@@ -44,12 +46,24 @@ export const usage: CommandUsage = () =>
...
   })
 
-export const views: Views<unknown> = {
-  defaultView: 'json',
+export const views = (
+  res: string,
+  _: unknown,
+  conf: LoadedConfig,
+): string => {
+  if (conf.positionals[0] === 'init') {
```
lukekarrys, line 54: if `pkg init --view=json` is run should it return json? all the other subcommands return json by default. i think by default `pkg init` can return the human readable text but i think json should be an option

src/git/src/user.ts
```diff
@@ -0,0 +1,64 @@
+import { spawn } from './spawn.js'
+
+export type GitUser = {
+  name?: string
+  email?: string
+}
+
+export const getUser = async (
```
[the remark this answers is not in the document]
ruyadorno: I like that idea 👍

src/vlt/src/commands/pkg.ts
ruyadorno: right, it's the one thing from `init` that is not fitting very well with the `vlt pkg` model, what do you think of having it just return json (the contents of the `package.json` file) in `vlt pkg init` but the more traditional human output in `vlt init` ? is that more confusing?

ruyadorno: btw, I really like having that extra info pointing users to use `vlt pkg set` to customize their `package.json` file at the end of the output:

```
vlt pkg init
Wrote to /Users/ruyadorno/tmp/test-foo/package.json:

{
  "name": "test-foo",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "author": "Ruy Adorno"
}

Modify & add package.json properties using `vlt pkg`, e.g:

  vlt pkg set "description=My new project"
```

lukekarrys (approved):

PR #379, opened 2025-02-05
fix: remove turbo & cli wrapper in lieu of `pnpm <bin>` fallback

darcyclarke:
### Description
- removes `turbo` as a dep (we weren't leveraging this & can revisit at a later date) as it was silently failing when run
- removes the `./scripts/local-cli` bin wrapper & `scripts` refs which are no longer required
- you can run `pnpm <vlt|vlr|vlx|...>` & it will fallback to the locally build bin (`pnpm run <bin>` will continue to not work)


lukekarrys:
We should remove the `local-cli` altogether if its not wrapping anything.

If we want `pnpm run vlt` to work we can change the run scripts to this:

```json
"vlt": "./node_modules/.bin/vlt"
```

Or we can delete them entirely because `pnpm` will look in `node_modules/.bin` automatically when no subcommand is given. So `pnpm vlt` will work without the run scripts, but `pnpm run vlt` won't.

lukekarrys (approved):

Issue #361, opened 2025-01-28
[FEATURE] GUI: Introduce table & card view switcher


darcyclarke:
### Description
Right now there is no easy way to introduce more sorting options to the list views (ex. querys & projects). Table components are ideal for this kind of sorting & often have this baked into their feature set.

### Suggestion
- [x] introduce a new view switcher (ie. card <> table)
- [x] introduce sorting view column value (ex. name, date, directory...)
- [x] introduce this option where it makes sense (ie. project listing vs. the query listing may not make sense to have a card view but should have a consistent table view experience)

PR #383, opened 2025-02-05
gui table view

ejkorol:
Creates a table view for the dashboard using [tanstack table](https://tanstack.com/table/latest)

fixes: #361

### Dark Mode
Screenshot 2025-02-05 at 17 15 19

### Light Mode
Screenshot 2025-02-05 at 17 12 00


src/gui/test/components/dashboard-grid/dashboard-sort-toggle.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+vi.mock('@/components/ui/toggle.jsx', () => ({
+  Toggle: 'gui-toggle',
+}))
+
+vi.mock('lucide-react', () => ({
+  ArrowDownAz: 'gui-arrow-down-az',
+  ArrowDownZa: 'gui-arrow-down-za',
+}))
+
```
ruyadorno, line 15:
```suggestion
```

I don't think those are needed

src/gui/src/components/dashboard-grid/dasboard-table.tsx
```diff
@@ -0,0 +1,117 @@
...
+    updateQuery(DEFAULT_QUERY)
+    updateActiveRoute('/explore')
+    updateStamp()
+  } else {
+    updateActiveRoute('/error')
+    updateErrorCause('Failed to select project.')
+  }
+}
+
+const DashboardTable = ({
```
ruyadorno, line 76:
```suggestion
export const DashboardTable = ({
```

I think the standard has been to export each component function directly, is there any particular reason for changing it to exporting an object ref at the end?

src/gui/test/components/dashboard-grid/__snapshots__/dashboard-table.tsx.snap
ruyadorno, line 1: I'd love if we could find a way to render these objects one day... anyways, fine for now

ruyadorno (changes requested):
A few things to clean up but otherwise it's looking great!

src/gui/src/components/dashboard-grid/dasboard-table.tsx
ruyadorno: if not, let's make sure to change all files to export the function instead 🙏 i noticed most components from this PR are like this

src/gui/src/components/data-table/data-table.tsx
```diff
@@ -0,0 +1,173 @@
...
+  filterValue,
+  setColumnVisibility,
+  columnVisibility,
+  onClickHandler,
+}: DataTableProps<TData, TValue>) => {
+  const [sorting, setSorting] = useState<SortingState>([])
+  const [globalFilter, setGlobalFilter] = useState<string>('')
+  const [pagination, setPagination] = useState<PaginationState>({
+    pageIndex: 0,
+    pageSize: 10,
```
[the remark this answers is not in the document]
ejkorol: I see what you mean! I think this will also benefit the `dashboard-view-toggle` to not always default back to the `grid` view... I'll make an issue for this an address both of these in a separate PR actually :)

PR #388, opened 2025-02-06
allow control regex because this is a CLI

lukekarrys:
From https://eslint.org/docs/latest/rules/no-control-regex

> Control characters are special, invisible characters in the ASCII range 0-31. These characters are rarely used in JavaScript strings so a regular expression containing elements that explicitly match these characters is most likely a mistake.

We're writing a CLI where these are not rarely used. And as a mistake it's unlikely and not worth the cost of ignoring it.


isaacs:
Let's land #373 first, and have this pick that one up as well.

isaacs (approved):

darcyclarke (approved):

PR #394, opened 2025-02-07
add pnpm engines

lukekarrys:


infra/build/src/bin/publish.ts
```diff
@@ -77,9 +77,14 @@ const PRERELEASE_ID = `.${Date.now()}`
...
   LICENSE: readFile(join(Paths.CLI, 'LICENSE')),
-  PACKAGE_JSON: readPackageJson(
-    join(Paths.CLI, 'package.json'),
-  ) as Bundle,
+  PACKAGE_JSON: (() => {
+    const pkg = readPackageJson(
+      join(Paths.CLI, 'package.json'),
+    ) as Bundle
+    // Only keep the node engines for publishing
+    pkg.engines = { node: pkg.engines.node }
```
darcyclarke, line 85: this doesn't seem to be working... I'm still seeing `engines.pnpm` references in the built package when doing a dry run (ie. `pnpm -F infra-build exec node dist/esm/bin/publish.js`)

darcyclarke (changes requested):

infra/build/src/bin/publish.ts
darcyclarke: *had to rebuild... looks like it's working now

darcyclarke (approved):

PR #412, opened 2025-02-10
dev: nave support and TS dev documentation

isaacs:


lukekarrys (approved):

CONTRIBUTING.md
```diff
@@ -29,3 +29,33 @@ These are websites that get deployed. Currently only [docs.vlt.sh](https://docs.
...
 and to attempt to fix them.
+
+## Running TypeScript Directly
+
+If you are on the latest version of Node, you can run TypeScript
+files directly using `node --conditions=@vltpkg/source
+./path/to/file.ts`.
+
+You can also `export NODE_OPTIONS=--conditions=@vltpkg/source` to
+set this in the environment, rather than on each command.
```
lukekarrys, line 40: Is it worth noting that the `export NODE_OPTIONS` would be preferred because that will propagate to place where we run `spawn(process.execPath)`? The places we do that are not load bearing but might cause some head scratching if `rollback-remove` failed for example.

isaacs: Oh, sure, maybe we should just make that the recommendation? I mean, it's mostly just for us and for when we recruit someone new.

isaacs: Updated to just say to do it this way.

lukekarrys (approved):

PR #413, opened 2025-02-10
dev: add root level scripts for local dev and building

lukekarrys:
Same idea as #412 but in the form of root level run-scripts that are helpful for people who run the CLI with `pnpm vlt` (eg #379).

This works because `pnpm <name>` will look in scripts before it looks in `node_modules/.bin`.

But to make this behavior unambiguous, the `node_modules/.bin/vlt` files that were written by pnpm have been removed. These were only used for local dev as the publish process rewrites `package.json#bin` to point to the correct files after bundling/compiling.

---

This PR also adds scripts that will run the various permutations of the scripts in `infra/build` to generate bundled, compiled or packed versions of the build for local testing.


darcyclarke (approved):
Thank you for doing this... I'm likely the `#1` person running `pnpm <bin>` atm

lukekarrys:
Converting to a draft to make one more change so that there's one preferred root script that has the default behavior we discussed @darcyclarke

PR #416, opened 2025-02-11
drop node 20 from CI, support

lukekarrys:
This will not pass the required status checks since those are matched by name.

Once this is merged I will update the names of the required status checks and current PRs will need to be rebased if they need to get required checks passing (or they can just bypass branch protections).


lukekarrys:
This will not pass the required status checks since those are matched by name.

Once this is merged I will update the names of the required status checks and current PRs will need to be rebased if they need to get required checks passing (or they can just bypass branch protections).

ruyadorno (approved):

PR #423, opened 2025-02-12
install: dont use tsx for now

lukekarrys:
This is a stop-gap to allow the CLI TypeScript source to be run completely with Node.

A future PR will revert this commit while also moving it to a new workspace that will require building in order to compile the JSX to JS.

I'll do this the next time I add a new `ink` based reporter.


isaacs (approved):
Ok, unpopular take maybe, but I kinda like the `$()` syntax better?

LGTM

lukekarrys:
I don't hate looking at it but I made myself type it out to get a feeling and made a few mistakes that are hard to make in JSX. Mainly the last argument of createElement being variadic but can also accept an array.

PR #427, opened 2025-02-13
test: run with node type stripping instead of ts-node

lukekarrys:
This configures tap to turn off the typescript plugin which uses ts-node
and instead run tests with node's type stripping flags.

When run this way, the tests no longer need tshy to be run during pretest
or presnap events. A future change could remove prepare events also but
for now it is still required for the gui and docs since those look for
the default export condition.

This requires using top level type imports instead of inline imports,
because as the the types are stripped inline type imports will leave
the import but with no bound values. This would be hard to cause an
actual bug because the imported file would have to cause side effects,
but it did cause coverage to be off for imports with side effects like
`xdg`.


lukekarrys:
Split into #435 and #436

PR #435, opened 2025-02-14
lint: prefer top level type imports

lukekarrys:
Type stripping is safer when we have separate imports for types and values.

I doubt it would cause bugs but I ran some of our tests and there were some coverage issues because an import like `import { type X } from 'package` gets stripped to `import {} from 'package'` but `import type { X } from 'package'` gets removed entirely.

We should probably be using TypeScript's [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) too/instead, but in turning it on it caused `pnpm -F gui prepare` to break. I'll probably look into this at some point in the future so the gui can also consume the raw TypeScript source.

So for now this is a lint rule only which has a similar effect. This PR only switches the lint rules from `prefer-inline` to `prefer-top-level` for the [consistent-type-specifier-style](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/consistent-type-specifier-style.md) rule and then runs `pnpm lint && pnpm format`.


ruyadorno (approved):
LGTM 👍

PR #436, opened 2025-02-14
test: run with node type stripping instead of ts-node

lukekarrys:
This configures tap to turn off the typescript plugin which uses ts-node and instead run tests with node's type stripping flags.

When run this way, the tests no longer need tshy to be run during pretest or presnap events. A future change could remove prepare events also but for now it is still required for the gui and docs since those look for the default export condition.


patches/@tapjs__test.patch
isaacs, line 1: Patches?? PATCHES?!? we don't need no stinkin patches

image

Just get Tap 21.1 and use `--type-strip-only` on the cli, or `type-strip-only: false` in the `.taprc` or `"tap": { "type-strip-only": false }` in the package.json

isaacs (changes requested):
tap can do this for us now, otherwise LGTM

lukekarrys:
The first commit in this PR is now rewritten to use `type-strip-only` instead of a patch.

PR #437, opened 2025-02-14
gui: queries should be always trimmed

ruyadorno:
Fixes an issue found when clicking on dependents.

Previously clicking on the root node would take you to:
`http://localhost:7017/explore?query=%3Aroot%20`

Now it takes you to:
`http://localhost:7017/explore?query=%3Aroot`


ejkorol (approved):
Good catch!

Issue #360, opened 2025-01-28
[BUG] GUI: Dependency search gets laggy when querying large projects


darcyclarke:
### Description
When writing queries against larger projects with hundreds or thousands of dependencies the typing/search experience can begin to lag & be slow to respond. It seems like the searching/filter is directly tied to keydown event which begins to hang the end-user experience & making it seem worse then it should be

### Suggestion
- [ ] implement a debounce or some kind of throttle which can wait for a period of time in between firing off search/filtering & improving the performance
- [ ] investigate alternative methods to binding the search typing & the filter experience so that the perceived end-user performance is improved

### Example (input lag starts being noticeable ~20s in with more greedy queries)

https://github.com/user-attachments/assets/67f42ef2-febe-4cba-bcf9-39d8d29518d1

PR #441, opened 2025-02-14
gui: add ability to cancel ongoing query search

ruyadorno:
Improves the responsiveness of the UI by cancelling any ongoing query search prior to creating a new one.

Fixes: https://github.com/vltpkg/vltpkg/issues/360
Depends on: https://github.com/vltpkg/vltpkg/pull/440

### DIFF
--- a/src/gui/src/app/explorer.tsx
+++ b/src/gui/src/app/explorer.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 import { Query } from '@vltpkg/query'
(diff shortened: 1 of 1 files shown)


lukekarrys:
Change looks good but I'm wondering if this should close #360. The issue we were seeing that @darcy wrote up was that running a single query could take awhile and freeze the UI even if no other query was already in transit.

lukekarrys:
I haven't this change yet but I will on Monday. That comment was just from a first glance at the code.

darcyclarke (approved):

Issue #150, opened 2024-09-26
[FEATURE] Binaries and Sourcemaps: what do they know? do they know things? Let's find out!


isaacs:
1. When compiling with ESBuild make sure that the compiled esbuilt code can sourcemap all the way back to the TS.
2. When compiling to Deno binary, same deal, get all the way back to the TS line/column numbers.
3. Maybe this means inlineSourceMaps? Try out some stuff.

Having useful error stack traces is definitely worth paying a little bit of a performance tax on.

lukekarrys:
Dropping a note here that deno binaries cannot use the source maps emitted by esbuild https://github.com/denoland/deno/issues/4499

lukekarrys:
We are currently blocked on distributing as a compiled binary for other reasons, but I have been able to get `esbuild` to generate sourcemaps all the way back to the original TypeScript. That coupled with using `#!/usr/bin/env -S node --no-warnings --enable-source-maps` as our hashbang means we are currently getting stack traces that map all the way back to the original TS (ref https://github.com/vltpkg/vltpkg/pull/737/files)

We have a separate tracking project in Linear (https://linear.app/vlt/project/build-single-executable-distribution-f231cdf9df9f/overview) for compiled binary distribution where I have added sourcemap parity as a todo once that is unblocked.

So I think this issue can be closed.

PR #444, opened 2025-02-15
build: run esbuild directly on typescript source

lukekarrys:
This also cleans up a few build related items:

- Uses esbuild's built-in code splitting. The tradeoff is that we have to put a few environment variables in the code. It's a bit of a pain but more explicit and less brittle than the previous approach.
- Uses esbuild's source maps which now map directly back to the TS source. Previously the sourcemaps were doing an extra hop from TS -> JS -> bundled JS. This makes stack traces in the bundled published code better and solves the non-compiled portion of #150.

Builds on #436


darcyclarke (approved):
Looks good, thanks for quickly fixing the compilation (looks like there's some conflicts but I think this is good to go once those are fixed)

PR #447, opened 2025-02-15
dev: remove custom source export condition

lukekarrys:
This is an experiment with a patched version of `tshy` to make the `default` export condition point directly at the TypeScript source and not actually build anything to `./dist`. This allows us to remove all the custom config tooling `conditions` and needing to set `--condition` in `NODE_OPTIONS`.

`tshy` will still do the right thing and build/export `./dist` files when run during workspace `pack` or `publish`. We're not actively publishing our workspaces but it's nice that this will Just Work when we do.


darcyclarke (approved):

lukekarrys:
Closing as this is part of #446 now.

Issue #261, opened 2024-12-11
[BUG] process stalls between command finish and `process.exit`


lukekarrys:
With the following changes to the vlt bin, I was measuring the time between the command finishing and `process.exit` being run.

When running `vlt i` and one of the deps fails to resolve, a bunch of active fs requests are left hanging which take time to clean up.

```patch
diff --git a/src/vlt/src/bins/vlt.ts b/src/vlt/src/bins/vlt.ts
index c9f614c2..53227b5b 100644
--- a/src/vlt/src/bins/vlt.ts
+++ b/src/vlt/src/bins/vlt.ts
@@ -1,2 +1,12 @@
 #!/usr/bin/env node
-void import('../index.js').then(r => r.default())
+let time = 0
+process.on('exit', () => {
+  console.log(`before exit ${Date.now() - time}ms`)
+})
+await import('../index.js')
+  .then(r => r.default())
+  .then(() => {
+    console.log('really really done!')
+    time = Date.now()
+    console.log((process as any)._getActiveRequests())
+  })
```

Here's an example of running with those changes on this package.json:

```json
  "dependencies": {
    "@astrojs/check": "^0.9.4",
    "@astrojs/mdx": "^3.1.9",
    "@astrojs/react": "^3.6.3",
    "@astrojs/starlight": "^0.29.2",
    "@astrojs/tailwind": "^5.1.3",
    "@astrojs/vercel": "^7.8.2",
    "@fontsource-variable/manrope": "^5.1.0",
    "@radix-ui/react-dialog": "^1.1.2",
    "@radix-ui/react-icons": "^1.3.0",
    "@radix-ui/react-separator": "^1.1.0",
    "@radix-ui/react-slot": "^1.1.0",
    "@radix-ui/react-tooltip": "^1.1.3",
    "@types/react": "^18.3.11",
    "@types/react-dom": "^18.3.1",
    "astro": "^4.16.17",
    "class-variance-authority": "^0.7.0",
    "clsx": "^2.1.1",
    "gray-matter": "^4.0.3",
    "lucide-react": "^0.453.0",
    "react": "^18.3.1",
    "react-dom": "^18.3.1",
    "sharp": "^0.33.5",
    "tailwind-merge": "^2.5.2",
    "tailwindcss": "^3.4.14",
    "tailwindcss-animate": "^1.0.7",
    "typedoc": "^0.27.4",
    "typedoc-plugin-frontmatter": "^1.1.0",
[...]

PR #449, opened 2025-02-15
cli: output/view refactoring and test coverage

isaacs:
This is a somewhat significant refactoring of the view/output machinery, that comes along with a fix for the hang identitfied in #261.

As of this change, cliCommand methods no longer return the result nested under a `result` key in the returned object. They just return the result as-is. If they return void, then that's what the return is. To return either a value or undefined, simply let the return value be something like `SomeType | undefined`.

This also fixes the strange inconsistency where the `vlt pkg` command would _always_ output JSON, even in `--view=human` mode. Now, _every_ command will support at minimum `human`, `json`, and `inspect` views. If `--view=inspect` is set, then the output object will be logged using `console.log()`.

If a view is specified that is not supported for the given command, then it will be output using the default view (`human` if stdout is a TTY, `json` if not).

Similarly, there's no longer a need to nest views under a `{ defaultView, views }` object, or pass around a `isJson` flag. We can just use the `config.values.view` as the one to use, and if that's `json`, then that's how we know it's JSON.

This is _not_ quite the permanent solution, because the ideal way to fix the hang is to pass an AbortSignal object through the tree of requests, so that it does not forcibly exit the process if for example an install is performed from the GUI. The AbortController/AbortSignal fix will come in the next commit.

Fix: #261


src/vlt/src/commands/list.ts
```diff
@@ -55,17 +52,20 @@ export const usage: CommandUsage = () =>
...
+  json: jsonOutput,
+  mermaid: mermaidOutput,
+  human: humanReadableOutput,
+  gui: async ({ queryString }, _, conf) => {
+    await startGUI({
+      conf,
+      startingRoute:
+        '/explore?query=' + encodeURIComponent(queryString),
+    })
   },
```
lukekarrys, line 67: Yes! This is what I had envisioned originally but couldn't get it play nicely at the time with output. But I think this is one example that shows this whole PR is the refactor in the right direction.

lukekarrys (approved):

ruyadorno (approved):

PR #450, opened 2025-02-18
vlt: mock file system lookup on start-gui tests

ruyadorno:
start-gui tests should not lookup the actual file system on its tests. Instead we should mock these internal calls to have a deterministic and faster way to run these tests in different systems.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -51,6 +51,9 @@ catalogs:
     prettier:
       specifier: ^3.4.2
       version: 3.4.2
+    promise-call-limit:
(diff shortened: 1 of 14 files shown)


isaacs:
Ok, this makes the test work fine.

But the actual `vlt gui` command still takes about 20 seconds to a minute to open a web browser on my machine, because it sits there waiting to read every project on my machine.

If this data is actually needed, it ought to be streamed in, or presented in an explorable way (like, a folder tree that expands and only loads what it finds at each level).

Otherwise, it feels a bit broken.

isaacs (approved):

PR #461, opened 2025-02-20
gui: ignore things from luke's home dir

lukekarrys:
`.config` and `.cache` are my `$XDG_CONFIG_HOME` and `$XDG_CACHE_HOME` respectively so it might be possible to use `@vltpkg/xdg` to ignore those. But this seemed easier as an interim solution. My goal is to get the default dashboard a bit more usable on my machine since it was detecting a lot of projects in these dirs.

**.vscode**
Screenshot 2025-02-19 at 11 27 02 PM

**.config**
Screenshot 2025-02-19 at 11 26 54 PM

**.cache**
Screenshot 2025-02-19 at 11 26 40 PM

### DIFF
--- a/src/vlt/src/ignored-homedir-folder-names.ts
+++ b/src/vlt/src/ignored-homedir-folder-names.ts
@@ -12,6 +12,9 @@ export const ignoredHomedirFolderNames = [
   'library',
   'desktop',
   'dropbox',
+  '.config',
+  '.cache',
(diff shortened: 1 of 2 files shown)


ruyadorno (changes requested):
the current logic should be ignoring any dot-folder, it looks like a bug that these would have been traversed

Issue #356, opened 2025-01-28
[FEATURE] `query`: Add `:semver` selector


darcyclarke:
### Description
Currently the only way to query dependencies by version is utilizing the attribute selector. There's a variety of usecases that are unlocked when you allow users to use the same semantic version range grammar that use with their dep specs in queries (ex. `>1`, `<1`, `~1.2.3` etc.).

### Suggestion
- [ ] implement the `:semver` pseudo selector in the dependency selector syntax
- [ ] consider diverging from `npm`'s [implementation](https://docs.npmjs.com/cli/v11/using-npm/dependency-selectors#semverspec-selector-function) & changing the order of arguments to create a more straightforward format (ie. `:semver(<spec>, [function], [selector])`).
  - notably, keeping the npm argument arrangement would require us to copy the "infer" logic which is less then desirable

PR #469, opened 2025-02-20
query: add semver pseudo selector

ruyadorno:
Add a new `:semver()` pseudo selector to the query language.

It supports 3 parameters:
- semver value: A valid semver version | range to compare. Supports quoted and unquoted strings.
- [function]: A semver function to use for the comparison, defaults to `satisfies`. Supports quoted and unquoted strings.
- [attribute]: An attribute selector to define a different `package.json` property to check, defaults to `[version]`. Supports one of:
  - attribute selector, e.g: `[version]`
  - :attr pseudo-selector, e.g: `:attr(engines, [node])`
  - string: `"version"`

### Examples:
```
  Selects packages that satisfies the `^2.0.0` range:
    :semver(^2.0.0)

  Selects package with a version different than provided:
    :semver("1.0.0", "neq")

  Selects packages that have a `engines.node` that satisfies `22.13.0`:
    :semver(22.13.0, satisfies, :attr(engines, [node]))
```
Fixes: https://github.com/vltpkg/vltpkg/issues/356


lukekarrys (approved):
Looks good!

I think we need more functions in `@vltpkg/semver` now. The engines example is most useful with `subset` which we don't have yet. We should follow-up this by adding `subset` (and maybe `intersects`) to `@vltpkg/semver` and then `:semver`.

PR #474, opened 2025-02-22
Isaacs/factor out gui server phase 1

isaacs:
This is phase 1 of the process of factoring out the gui-server for greater iteration, performance, and ease of development.

1. All gui-server methods now specifically define the options that they actually need, rather than passing around the `LoadedConfig` god-object (which will not be available to the extracted-out gui-server daemon).
2. The `LoadedConfig` god object is still used by the top-level `startGUI` method, but this will eventually be factored out as well. Currently, this is a bit tricky to do in place, because it does need to occasionally call `conf.resetConfig()` in a few scenarios. My thinking here is that we'll have `@vltpkg/gui-server` export an object type, and have a nice IOC way to listen for its requests to re-calculate configuration, and a method to provide these updates. Something like `const server = createServer({ ... some options ... }); server.on('resetConfig', folder => { conf.resetConfig(folder); server.updateConfig(conf.options) })`. Idk, this feels a bit verbose, probably can be tersed up a bit, but that's the general idea.
3. `init` is moved out to a new workspace `@vltpkg/init`.
4. `install()` and `uninstall()` are moved into `@vltpkg/graph` where they belong.

Phase 2, once this lands, will implement the `@vltpkg/gui-server` workspace, which will then take the `ignored-homedir-folder-name.ts`, `gui-server.ts`, `project-info.ts`, and `read-project-folders.ts` modules away from the cli, and export an interface for the cli to create and communicate with a gui server.

Phase 3, once that is done, will be the point where we can start iterating on the gui-server functionality, replacing the "run a brand new stateless server on each invocation" with a sock-daemon that can run in the background, update its configs as needed, perform various actions when told to by the CLI or the gui interface, sync from a remote server, and so on, all without blocking the main process or having to hang for its initial data setup on each command.


lukekarrys (commented):
Code and description look good to me.

Tests are failing with a new timeout in the graph workspace for `test/reify/index.ts`. I was able to reproduce locally with `pnpm -F graph test -- test/reify/index.ts`. So once CI is green this is good to land.

src/vlt/README.md
```diff
@@ -7,3 +7,42 @@ The command line interface for `vlt`.
...
+In software design pattern terms, this is a "View Controller". It
+parses the user's intent, passes functionality off to some other
+module, and then presents the result in the appropriate format.
+It does _not_ implement that functionality or contain other
+opinions unrelated to "what is the user asking for" and "how to
+present the result".
+
+Basically, if it's not "parse user intent" or "print the result",
+then it belongs elsewhere! You can spike it out here, but it
+_must_ find a home in another project.
```
lukekarrys, line 48: Using this comment as a reminder to myself to make an explicit README.md in `infra/build` instead of copying this one. This is the right place for documentation like this, but probably not valuable in the published package.

lukekarrys: @isaacs Actually what do you think about putting this in the CONTRIBUTING.md?

isaacs: @lukekarrys What do you think of this approach? 396864aa27676b219de44a06c40be9321115549b

PR #480, opened 2025-02-25
gui: updates metadata styles

ejkorol:
<details>
<summary>
<h3>After:</h3>
Screenshot 2025-02-25 at 19 35 02
</summary>
<h3>Before:</h3>
Screenshot 2025-02-21 at 13 48 26
</details>

<details>
<summary>
<h3>Video Overview!</h3>
</summary>

https://github.com/user-attachments/assets/b2c802ec-2308-4971-896e-4bb52c4aaec6
</details>

Closes vltpkg/statusboard#75
- updates basic styling on `selected-item` component
- replaces metadata icon with custom `png`
 - updates `scripts/prepare.js` to load images
- adds `tremor` chart dependencies
 - installs recharts
 - adds `src/lib/chart-utils.ts` boilerplate
 - adds tremor inputs in `src/lib/utils.ts`
 - adds `spark-chart.tsx` tremor ui component
- adds shadcn component `scroll-area.tsx`
 - installs `@radix-ui/scroll-area` dependency
- updates `shiki` codeblock theme
- updates styling on `explorer-grid` components
- adds download graph on explorer-grids `selected-item`
 - updates `src/lib/external-info.ts` to fetch yearly download count
 - adds `src/utils/transform-weekly-downloads.ts` to transform yearly
   download count
 - adds `spark-chart` into `selected-item` component
 - updates tests and snapshots


darcyclarke (changes requested):
Looks like you're using two different pinks for fonts, one for the spec & another for the json keys - which is just off enough that it feels like a mistake/makes my eyes bug out (see attached)

Screenshot 2025-02-25 at 6 52 25 PM

src/gui/src/lib/chart-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+  maxValue: number | undefined,
+) => {
+  const minDomain = autoMinValue ? 'auto' : (minValue ?? 0)
+  const maxDomain = maxValue ?? 'auto'
+  return [minDomain, maxDomain]
+}
+
+// Tremor Raw hasOnlyOneValueForKey [v0.1.0]
+
+export function hasOnlyOneValueForKey(
```
lukekarrys, line 106: Is this function being used anywhere?

lukekarrys: If not, I think we should remove it. I did check the tremor docs and their description of what the function does seems incorrect.

From https://tremor.so/docs/utilities/chartUtils#hasonlyonevalueforkey
> checks if all objects in an array have the same value for a specified key. If all objects have the same value for the given key, it returns true; otherwise, it returns false.

But that's not what's happening in the code:

```js
console.log(hasOnlyOneValueForKey([{ a: 1 }], 'a'))
// true

console.log(hasOnlyOneValueForKey([{ a: 1 }, { a: 1 }], 'a'))
// false
```

src/gui/src/components/ui/shiki.tsx
```diff
@@ -28,10 +33,929 @@ export function CodeBlock({
...
+          )}>
           <div dangerouslySetInnerHTML={{ __html: out }} />
         </div>
       : ''}
     </>
   )
 }
+
+const customDarkTheme = `{
+  "name": "Vercel Theme",
```
lukekarrys, line 51: Is this name used fore display anywhere? If so we should change it.

src/gui/scripts/prepare.js
```diff
@@ -43,6 +43,9 @@ const js = async () => {
     define: {
       'process.env.NODE_ENV': `"${argv.production ? 'production' : 'development'}"`,
     },
+    loader: {
+      '.png': 'file',
+    },
```
lukekarrys, line 48: We don't need to add this unless we use `import` for the images which I didn't see in the code.

Everything in `public/` is copied to `dist/`, so if we only reference them as strings in `src` attributes as is done in this PR, we can remove the loader.

src/gui/src/components/ui/shiki.tsx
lukekarrys: If it's not, then not a big deal but wanted to flag just in case.

ejkorol: It is not being used for display, I believe it was just the identifier from the VSCode theme (I pulled the same file from the Docs workspace) I will change it tho :)

src/gui/src/lib/chart-utils.ts
ejkorol: I dont believe it is used specifically in our code here, so I will remove it for this PR, but it may reappear later on should we continue to use other tremor components (such as LineChart etc):

https://github.com/tremorlabs/tremor/blob/main/src/components/LineChart/LineChart.tsx#L559

lukekarrys (approved):

PR #493, opened 2025-03-05
docs: style updates and improvements

### LINKED ISSUE #287: docs: dont use `client:only="react"` for sidebar
This makes the sidebar not load at all if JS is slow/fails. I think it would be better to always render the sidebar and rely on client JS to set scroll position/collapsed state after it is visible.

### LINKED ISSUE #297: docs: audit use of client:only="react"
Ideally for a content driven site like docs, we wouldn't need to use this at all. I'm not sure if it's viable to go that far, but we should at least audit our usage of `client:only` and see if there are cases where we can progressively enhance instead with a good default.

ejkorol:
A necessary improvement and enhancement of the docs

### Before
Screenshot 2025-03-05 at 13 11 50

### After
Screenshot 2025-03-05 at 13 12 09

Closes #297 
Fixes #287 
Closes vltpkg/statusboard#68
Fixes vltpkg/statusboard#81

- **feat: adds new icons, removes unused / old ones**
- **chore: cleans up astro component overrides**
- **feat: redesigns sidebar**
- **chore: cleans up `page-sidebar` component and styles**
- **chore: cleans up unusued logic in linear-menu**
- **feat: upates root layout in `page-frame.astro`**
- **feat: creates `hero` section in `.mdx`**
- **chore: updates `globals.css` organization and styles**
- **feat: creates `selectors` page for dependency selector docs**
- **chore: runs lint and format**


lukekarrys (approved):
This looks great!

I noticed a small issue with the sidebar where you can't close an open dropdown without also navigating to that page.

But I think this is a huge improvement that we should merge and log/fix that issue in a follow-up.

PR #497, opened 2025-03-06
gui: query creation

ejkorol:
Allows a user to create a new Query from the Queries view, also adds some small animation based interactions and some styling updates.

Closes vltpkg/statusboard#80
Fixes vltpkg/statusboard#79

https://github.com/user-attachments/assets/57c826fb-e055-47c4-bf33-875a740ab07e

#### Note:
The last commit of this PR is a small chore that prefers the UI to use `cursor-default` instead of `cursor-pointer` whenever it is not immediately obvious that an element allows for user-based interactions (i.e. clicks).

This aligns more with a 'native' design philosophy, and as a 'web app' we should close to native as possible.

We should focus on interaction cues that are primarily communicated through visual and motion based affordances such as press animations or hover states, rather than cursor changes to align more with the native philosophy

#### Changes:
- **bugfix**: invalid nesting of `button` element on `color-picker`
- **feat**: adds micro-interactions to `create-labels`
- **feat**: create query
- **bugfix**: sets a height on `header`
- **chore**: prefer to use `cursor-default` than `cursor-pointer`


ruyadorno (approved):
LGTM 👍

PR #509, opened 2025-03-07
docs: add using with vsr guide

ruyadorno:
Add a guide showing how to use a VSR instance with the vlt client.

### DIFF
--- a/www/docs/src/styles/globals.css
+++ b/www/docs/src/styles/globals.css
@@ -195,6 +195,7 @@ html {
 
 .sl-markdown-content h3 {
   font-size: 1rem;
+  padding-top: 1rem;
 }
 
 .sl-markdown-content p {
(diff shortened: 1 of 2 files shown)


www/docs/src/styles/globals.css
ruyadorno, line 1: There are a few tweaks to make sure the `Code` component now renders well for a file-like example:

Screenshot 2025-03-07 at 11 04 39 AM

cc @ejkorol

www/docs/src/styles/globals.css
```diff
@@ -195,6 +195,7 @@ html {
 
 .sl-markdown-content h3 {
   font-size: 1rem;
+  padding-top: 1rem;
```
ruyadorno, line 198: Here I'm giving some breathing space for `h3` tags as I find that everything is super packed and makes it for a difficult to read docs in general when you're using `h3` more often.

ejkorol (approved):

darcyclarke (approved):

PR #510, opened 2025-03-07
gui: update color styles

ejkorol:
Updates the background color styles on the gui

Closes vltpkg/statusboard#82
- It goes through all the files that set an explicit `bg-<color>` and sets a default for consistency
- It sets a `max-h` on the content
- Updates the color variables set in `main.css`
- Updates the `max-w` of the content to be slightly larger, and `text-sm` to be slightly smaller in `tailwind.config`
- Updates `bg` and `hover` states on radix/shadcn components that set an explicit `bg-<color>`

### Before
before

### After
after


src/gui/src/app/explorer.tsx
```diff
@@ -141,9 +141,9 @@ const ExplorerContent = () => {
   }
 
   return (
-    <section className="flex w-full grow flex-col justify-between rounded-lg bg-white dark:bg-black">
-      <div className="flex w-full border-[1px] px-8 py-4">
-        <div className="mx-auto flex w-full max-w-7xl items-center justify-between">
+    <section className="flex h-full max-h-[calc(100svh-65px-16px)] w-full grow flex-col justify-between overflow-y-auto rounded-b-lg border-[1px]">
```
ruyadorno, line 144: what is going on with max-h here?

ejkorol: For the list content to be scrollable, we limit the height for the overflow by setting a max-h for that content area, which in this case would be: `screenHeight - headerComponent - sidebarInsetMargin`

ruyadorno: 👍

ruyadorno (approved):

PR #512, opened 2025-03-07
gui: update directory select logic

ejkorol:
Replaces the old directory select logic, which was previous just an input field with a command/combobox that allows a user to select the directory from a list.

This PR creates a new `directory-select` component to reuse whenever the UI requires it.

Closes vltpkg/statusboard#84

### Before
Screenshot 2025-03-07 at 15 30 54

### After
Screenshot 2025-03-14 at 12 51 52

Screenshot 2025-03-14 at 12 52 00

- **feat**: creates `directory-select` component

- removes old path select, replaces with directory-select component
- updates queries saved-item with new `directory-select` comp
- updates `queries` view to fetch dashboard data
- updates directory select logic for new `directory-select` comp
- fetches dashboardData to pass into `create-query` comp
- chore: lint & format

### DIFF
--- a/src/gui/src/components/queries/queries-empty-state.tsx
+++ b/src/gui/src/components/queries/queries-empty-state.tsx
@@ -3,8 +3,13 @@ import { Button } from '@/components/ui/button.jsx'
 import { Command, Plus, Star } from 'lucide-react'
 import { AnimatePresence, motion, useAnimate } from 'framer-motion'
 import { CreateQuery } from '@/components/queries/create-query.jsx'
+import type { DashboardData } from '@/state/types.js'
 
-const QueriesEmptyState = () => {
+interface QueriesEmptyStateProps {
+  dashboard?: DashboardData
+}
+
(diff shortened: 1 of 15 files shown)


src/gui/src/components/queries/queries-empty-state.tsx
```diff
@@ -3,12 +3,42 @@ import { Button } from '@/components/ui/button.jsx'
...
+  )
+  const stamp = useGraphStore(state => state.stamp)
+
+  useEffect(() => {
+    startDashboardData({
+      updateActiveRoute,
+      updateDashboard,
+      updateErrorCause,
+      stamp,
+    })
```
ruyadorno, line 32: Let's make sure to pass down from the `app/queries.tsx` component that already have this data.

ruyadorno (changes requested):
Looking good 👍

ruyadorno (approved):
LGTM 👍

PR #518, opened 2025-03-12
security-archive: add a security db & cache system

ruyadorno:
Add a new package `@vltpkg/security-archive` that handles accessing, fetching remote data and persisting security report data on packages from a given graph.

## Usage

```js
import { actual } from '@vltpkg/graph'
import { SecurityArchive } from '@vltpkg/security-archive'
const specOptions = {
  registry: 'https://registry.npmjs.org/',
}
const graph = actual.load({
  ...specOptions,
  projectRoot: process.cwd(),
})
const archive = await SecurityArchive.start({ graph, specOptions })
if (archive.ok) {
  for (const node of graph.nodes.values()) {
    const securityData = archive.get(node.id)
    if (securityData) {
      console.log('securityData', securityData)
    }
  }
} else {
  console.warn('Failed to start the SecurityArchive')
}
```

### DIFF
--- a/src/security-archive/src/index.ts
+++ b/src/security-archive/src/index.ts
@@ -1,0 +1,385 @@
+// eslint-disable-next-line import/no-unresolved
+import { DatabaseSync } from 'node:sqlite'
+import { LRUCache } from 'lru-cache'
+import pRetry, { AbortError } from 'p-retry'
+import { loadPackageJson } from 'package-json-from-dist'
+import { asDepID, splitDepID, joinDepIDTuple } from '@vltpkg/dep-id'
+import { error } from '@vltpkg/error-cause'
+import { XDG } from '@vltpkg/xdg'
+import { asPackageReportData } from './types.ts'
+import type { DepID } from '@vltpkg/dep-id'
+import type { GraphLike } from '@vltpkg/graph'
+import type { SpecOptions } from '@vltpkg/spec'
+import type {
+  PackageReportData,
+  SecurityArchiveLike,
+  SecurityArchiveRefreshOptions,
+} from './types.ts'
+
+export type * from './types.ts'
+
+const SOCKET_API_V0_URL = 'https://api.socket.dev/v0/purl?alerts=true'
+const SOCKET_PUBLIC_API_TOKEN =
+  'sktsec_t_--RAN5U4ivauy4w37-6aoKyYPDt5ZbaT5JBVMqiwKo_api'
+
+export const targetSecurityRegisty = 'https://registry.npmjs.org/'
+
+export type JSONItemResponse = {
+  namespace?: `@{string}`
+  name: string
+  version: string
+}
+
+export type DBReadEntry = {
+  depID: string
+  report: string
+  start: number
+  ttl: number
+}
+
+export type DBWriteEntry = [string, string, number, number]
+
+export type SecurityArchiveOptions = LRUCache.OptionsBase<
+  DepID,
+  PackageReportData,
+  unknown
+> & {
+  /**
+   * Security archive does not supports a fetch-on-demand model.
+   */
+  fetchMethod?: undefined
+  /**
+   * An optional value for the path in which to store the sqlite db.
+   */
+  path?: string
+  /**
+   * Number of retries attempts to reach the remote security API.
+   */
+  retries?: number
+}
+
+/**
+ * Only the public npm registry has information available in the
+ * socket.dev API. This function checks if a given depID is pointing
+ * to the public npm registry and returns `true` if it does.
+ */
+const usesTargetRegistry = (
+  depID: DepID,
+  specOptions: SpecOptions,
+): boolean => {
+  const [nodeType, nodeRegistry] = splitDepID(depID)
+
+  if (nodeType !== 'registry') {
+    return false
+  }
+
+  const reg =
+    nodeRegistry === '' ?
+      specOptions.registry
+    : specOptions.registries?.[nodeRegistry]
+  return reg === targetSecurityRegisty
+}
+
+// Loads the version number to be used in the User-Agent header
+export const { version } = loadPackageJson(
+  import.meta.filename,
+  process.env.__VLT_INTERNAL_CLI_PACKAGE_JSON,
+) as {
+  version: string
+}
+
+/**
+ * A database of security information for given packages in a graph.
+ *
+ * Using the SecurityArchive.refresh() method will update the local cache
+ * with information from the socket.dev APIs or load from the local storage
+ * if available. Information about package security is then available
+ * using the SecurityArchive.get() method.
+ */
+export class SecurityArchive
+  extends LRUCache<DepID, PackageReportData>
+  implements SecurityArchiveLike
+{
+  #fetchedDepIDs = new Set<DepID>()
+  #path: string
+  #retries: number
+
+  /**
+   * True if the refresh process was successful and report data
+   * is available for all public registry packages in the graph.
+   */
+  ok = false
+
+  /**
+   * Creates a new security archive instance and starts the refresh process.
+   */
+  static async start(
+    options: SecurityArchiveOptions & SecurityArchiveRefreshOptions,
+  ) {
+    const archive = new SecurityArchive(options)
+    await archive.refresh(options)
+    return archive
+  }
+
+  /**
+   * By default, limits to 100K entries in the in-memory archive.
+   */
+  static get defaultMax() {
+    return 100_000
+  }
+
+  /**
+   * By default, entries are cached for 3 hours.
+   */
+  static get defaultTtl() {
+    return 1000 * 60 * 60 * 3
+  }
+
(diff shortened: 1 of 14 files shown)


src/security-archive/src/index.ts
```diff
@@ -0,0 +1,291 @@
...
+    nodeRegistry === '' ?
+      specOptions.registry
+    : specOptions.registries?.[nodeRegistry]
+  return reg === targetSecurityRegisty
+}
+
+// Loads the version number to be used in the User-Agent header
+const { version } = loadPackageJson(import.meta.filename) as {
+  version: string
+}
```
ruyadorno, line 68: Need to use the special env var here for builds, e.g:
```suggestion
export const { version } = loadPackageJson(
  import.meta.filename,
  process.env.__VLT_INTERNAL_CLI_PACKAGE_JSON,
) as {
  version: string
}
```

src/security-archive/src/index.ts
```diff
@@ -0,0 +1,297 @@
...
+  /**
+   * Store current in-memory items to the database.
+   */
+  #storeItemsToDatabase(db: DatabaseSync): void {
+    const entries = this.dump()
+    let insertData = ''
+    for (const [key, value] of entries) {
+      insertData += `('${key}', '${JSON.stringify(value)}'),`
+    }
+    insertData = insertData.substring(0, insertData.length - 1)
```
isaacs, line 251: This feels a bit clunky with the repeated appends and then substring at the end?

Consider:

```ts
const insertData: string[] = []
for (const [key, value] of entries) {
  insertData.push(`('${key}', '${JSON.stringify(value)}')`)
}
const dbWrite = db.prepare(
  `INSERT OR REPLACE INTO cache (key, value) VALUES ${insertData.join(',')}`,
)
dbWrite.run()
```

src/security-archive/src/index.ts
```diff
@@ -0,0 +1,297 @@
...
+  version: string
+}
+
+export type LoadedEntry = {
+  ttl: LRUCache.Milliseconds
+  start: LRUCache.Milliseconds
+  value: PackageReportData
+}
+
+export type SecurityArchiveOptions = {
```
isaacs, line 36: This should extend `LRUOptions`, so that we can set the max, ttl, etc.

If anything really should not be set, then you can make it only settable to `undefined` in this type, like `fetchMethod?: undefined`.

src/security-archive/src/index.ts
```diff
@@ -0,0 +1,297 @@
...
+   * By default, entries are cached for 3 hour.
+   */
+  static get defaultTtl() {
+    return 1000 * 60 * 60 * 3
+  }
+
+  constructor(options: SecurityArchiveOptions = {}) {
+    super({
+      max: SecurityArchive.defaultMax,
+      ttl: SecurityArchive.defaultTtl,
```
isaacs, line 115:
```suggestion
      ttl: SecurityArchive.defaultTtl,
      ...options,
```

src/security-archive/src/index.ts
```diff
@@ -0,0 +1,297 @@
...
+/**
+ * A database of security information for given packages in a graph.
+ *
+ * Using the SecurityArchive.refresh() method will update the local cache
+ * with information from the socket.dev APIs or load from the local storage
+ * if available. Information about package security is then available
+ * using the SecurityArchive.get() method.
+ */
+export class SecurityArchive
+  extends LRUCache<DepID, PackageReportData>
```
isaacs, line 82: If I'm reading this right, it seems like it loads _every_ security notice we've ever seen, every time it's used?

It seems like you could get better overall performance by leveraging the `fetchMethod` LRUCache behavior. That could make this more efficient in some cases, by avoiding loading data we don't end up needing.

src/security-archive/src/index.ts
```diff
@@ -0,0 +1,297 @@
...
+    }[]) {
+      const { start, ttl, value } = JSON.parse(
+        entry.value,
+      ) as LoadedEntry
+      this.set(asDepID(entry.key), asPackageReportData(value), {
+        ttl,
+        start,
+      })
+    }
+    this.purgeStale()
```
isaacs, line 158: Why is this needed? Can we avoid it by just not adding anything that we know is past its TTL?

src/security-archive/src/index.ts
```diff
@@ -0,0 +1,297 @@
...
+        )
+      }
+    }
+  }
+
+  /**
+   * Store current in-memory items to the database.
+   */
+  #storeItemsToDatabase(db: DatabaseSync): void {
+    const entries = this.dump()
```
isaacs, line 246: This can be quite a lot of stuff, most of which hasn't changed.

It'd probably be better to gather up a batch of changed items, and only write back to the DB periodically, as items are updated.

src/security-archive/src/index.ts
ruyadorno: of course!

isaacs (dismissed):
This is a pretty good first pass, but I'd probably want to think through the data flows a bit more carefully to minimize massive load/save cycles, while still taking advantage of being able to batch up all the requests for a set of nodes in a graph. Especially, `LRUCache.load()` and `LRUCache.dump()` are pretty inefficient, and can get slow when the data grows.

I'd approach this as sort of a 3-level system, where you have the in-memory LRU, backed up by the database, backed further by the Socket API.

So, at initial load time, you give it a set of node IDs, and it tries to load those from the database. Any that it doesn't find, or where the TTL is expired, it fetches from the API server (perhaps in the background, and then saves back to the DB when it returns), so there's effectively zero delay, and never a full cache load/dump cycle.

src/security-archive/src/index.ts
ruyadorno: currently it loads every security notice for the set of nodes found in the `graph` passed in the `refresh` method.

src/security-archive/src/index.ts
ruyadorno: how could we avoid not adding anything that we know is past TTL? I think that's a better solution 👍

src/security-archive/src/index.ts
ruyadorno: I don't think `fetchMethod` is going to be the best tool in the box for this particular task at hand. We do want to leverage the batch API Request, reading all packages all at once fits better that same model.

One more limitation / design goal that I'm going for in this current implementation that might not be straightforward is the fact that the SocketAPI can only be called from the backend (mostly due to a lack of CORS header in the server response) - that makes it so that the isomorphic implementation (at `./src/browser.ts`) can't really rely on a persistency + fetch API model that is available on node/backend.

The solution on my design is having the isomorphic implementation be a [standard map](https://github.com/vltpkg/vltpkg/pull/518/files#diff-df31307ba612f9fe6af94c88acf8ab099e483c379ac09cb781bc017be5763524R34-R37) that gets populated with data as part of the initial [`TransferData`](https://github.com/vltpkg/vltpkg/blob/main/src/gui/src/state/types.ts#L55-L59) (pratically speaking, this is the `graph.json` returned by the GUI server). It will make a snappier end-user experience in the frontend (which matters a lot since users are expected to be tweaking the search bar a lot) while having a tradeoff of having to load upfront info on all packages for that given graph, which again, in the frontend the likeness that users are going to tweak the searchbar to `*`, `:root > *` is very high, so having that information upfront makes sense.

src/security-archive/src/index.ts
ruyadorno: I'll go ahead and add that suggestion to the implementation to be explicit about not supporting `fetchMethod`.

src/security-archive/src/index.ts
ruyadorno: right, we could tweak the implementation to hold a set of entries that were actually fetched from the API so that only those need to be updated.

ruyadorno:
> I'd approach this as sort of a 3-level system, where you have the in-memory LRU, backed up by the database, backed further by the Socket API.
> 
> So, at initial load time, you give it a set of node IDs, and it tries to load those from the database. Any that it doesn't find, or where the TTL is expired, it fetches from the API server (perhaps in the background, and then saves back to the DB when it returns), so there's effectively zero delay, and never a full cache load/dump cycle.

You just described the current implementation, so I'll take that as a nod of approval 😄 I think the only point of contention from your initial review is the idea of maybe leveraging `LRUCache.fetch` to have an ondemand load experience and I tried to explain why I avoided that in [this comment here](https://github.com/vltpkg/vltpkg/pull/518#discussion_r1996197735).

src/security-archive/src/index.ts
ruyadorno: done ✅

ruyadorno:
Added since the first pass:

- remodelled the db, separating `ttl` and `start` values as their own db column
- when selecting from the db, filter out expired entries
- validation that all packages required as part of the graph have their reports ready (`archive.ok` property)
- added a basic retry logic using `p-retry`

lukekarrys (approved):

PR #527, opened 2025-03-17
fix: re-add desktop to the available locations for projects

darcyclarke:


ruyadorno:
cc @isaacs will it break in your current setup when using the GUI?

darcyclarke:
I keep a bunch of projects on my desktop. Seeing this location added/removed/added/removed is getting really frustrating/annoying. If there's a problem with this location/accommodation then I've got serious questions/concerns.

isaacs:
We need to prioritize a "add folders to the dashboard" workflow, and make that the initial experience.

There is no universal set of default folders that can be added that won't cause problems, either by virtue of being present or absent.

So, sure, add it back. But we need to get away from this paradigm entirely asap, in my opinion.

For myself, I've configured it to only search in a few folders where I keep node projects. But we can't rely on users knowing to do that. But yes, I do have a bunch of backed up Photo.app libraries and home directories in a folder under ~/Desktop, and several applications, all of which take several seconds each to fruitlessly scan for node projects. `~/Desktop` was actually the second worst offender, after `~/Documents`, because that's where VirtualBox stores its virtual machine images.

PR #538, opened 2025-03-18
Isaacs/integrity checking

isaacs:
check integrities


package.json
```diff
@@ -39,7 +39,7 @@
     "build:bundle": "vlt-build --outdir=.build-bundle bundle",
     "build:compile": "vlt-build --outdir=.build-compile compile",
     "deps:check": "node --no-warnings --experimental-strip-types scripts/check-deps.js",
-    "fix": "pnpm fix:pkg && pnpm lint && pnpm format",
+    "fix": "nave use 22 bash -c 'pnpm fix:pkg && pnpm lint && pnpm format'",
```
ruyadorno, line 42: won't this break folks that don't have a `nave` bin in their system?

lukekarrys: I opened https://github.com/vltpkg/vltpkg/pull/540 which should fix the behavior where running `fix` in 22 vs 23 would give different results. Once that lands, running linting with 23 should not be required anymore.

ruyadorno: so we can remove this change?

```suggestion
    "fix": "pnpm fix:pkg && pnpm lint && pnpm format",
```

lukekarrys (approved):

PR #542, opened 2025-03-19
compile: use deno --unstable flags instead of rolling our own

lukekarrys:
Previously `esbuild` was used for setting globals and converting to `node:` prefixed imports so that `deno compile` would work. This PR opts in to using deno's `--unstable-node-globals` and `--unstable-bare-node-builtins` flags since they do the same thing.


lukekarrys:
This might fix the error `fastNowTimeout.refresh is not a function` since we are no longer creating our own `setTimeout` global, but I could not reproduce it locally.

ruyadorno (approved):

PR #543, opened 2025-03-19
fix: `nvm` doesn't understand `"^"` `node-semver` range selectors

darcyclarke:
`nvm` does not understand all of npm-style semver range selectors (specifically `"^"`). The current range `^22.14.0` fails to be parsed & errors accordingly (although this _does_ work in CI because `setup-node` manually reads from & then evaluates this value using `node-semver`). This PR switches the value back to something `nvm` understands while trying to stay as close to the intent of the original change.

- switch nvm to use `lts/jod` which is the "latest" `22` release line
- switch our workflows to use `node-version` config which _does_ support npm-style semver range selectors
- document these changes in `CONTRIBUTING.md`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Setup Nodejs
         uses: actions/setup-node@v4
         with:
-          node-version-file: '.nvmrc'
+          node-version: '^22.14.0'
           cache: pnpm
           check-latest: true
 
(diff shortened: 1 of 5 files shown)


lukekarrys:
It was bumped to 22.14 because there was a bugfix in that version for experimental strip types. Any version  below that will not work for development.

darcyclarke:
I think we have a couple more options:

3. set/pin the value to the minor `22.14` (will still pick up patches although those are unlikely/minimal)
4. set the value to the specific `22` release line dist tag - ie. `lts/jod` (will pick up whatever the latest minor is - ref. https://github.com/nodejs/Release#release-schedule)

PR #549, opened 2025-03-19
query: add malware selector

ruyadorno:
Add a `:malware(<kind>)` selector that filters nodes based on malware-related alert types from the security-archive.


darcyclarke (changes requested):
Can you make the docs readme up-to-date with the new query selectors? (ref. https://github.com/vltpkg/vltpkg/blob/main/src/query/README.md) It was copied over from the query lib & I think maybe the query lib should just link to our docs & we can maintain just that one markdown doc with all the examples/selector references.

lukekarrys (approved):

PR #550, opened 2025-03-19
set globals from 'node:timers' in bundled code

lukekarrys:
It seems like Deno will sometimes use its [web timers](https://docs.deno.com/api/web/~/setTimeout) for `setTimeout`. I could only reproduce this when running the bundled code and not when running `deno` against the TypeScript source files. This PR fixes the issue by explicitly setting all the global timer variables from `node:timers` in bundled code.

This PR has also commits for:

- publishing `@vltpkg/cli-js` which is just the esbuild bundled JS for comparison testing 
- a smoke test for installing a package from `registry.npmjs.org`
- prefixes all the node builtin imports with `node:` so `deno` wont show warnings when debugging
- creates more directories inside `scripts/bins` for other variants of the executables: bundled node, bundled deno, and compiled deno


eslint.config.mjs
```diff
@@ -1,8 +1,8 @@
 import eslint from '@eslint/js'
 import tseslint from 'typescript-eslint'
 import globals from 'globals'
-import { readFileSync } from 'fs'
-import { resolve } from 'path'
+import { readFileSync } from 'node:fs'
```
isaacs, line 4: Is there an eslint rule we could apply/create to enforce this? Seems like a thing that we'll forget to do, like, *immediately*.

isaacs (approved):
Suggestion to make the `node:` prefixes (and maybe also `setTimeout` coming from `node:timers`) an eslint rule if possible, but LGTM.

PR #553, opened 2025-03-20
lk/deno cache unzip

lukekarrys:
- **publish: publish @vltpkg/cli-js for comparison testing**
- **install: failing smoke test for compiled bin install**
- **use node: prefix for all node builtins**
- **dev: add local deno bins for debugging from source**
- **fixup! dev: add local deno bins for debugging from source**
- **cache-unzip: pass correct flags when run in deno**


isaacs:
Looks like this needs rebase on main?

isaacs (commented):
LGTM once rebased on main.

lukekarrys:
I rebased thisn since it was built on some commits from #550. But it's not working with the compiled binaries because `process.execPath` points to the `vlt` bin and not `deno`. This isn't an issue with the runtime, but there needs to be a different way to spawn a process that will call the cache-unzip entry.

src/cache-unzip/src/unzip.ts
```diff
@@ -143,15 +141,24 @@ export const main = async (
...
         cache.set(key, Buffer.concat(chunks, headLength + unz.length))
       }
+      return true
     }),
   )
   await cache.promise()
-  // TS mistakenly things didSomething is always false
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-  if (!didSomething) process.exit(1)
+  if (!didSomething.some(Boolean)) process.exit(1)
```
isaacs, line 148: Cute.

src/cache-unzip/src/unzip.ts
```diff
@@ -143,15 +141,24 @@ export const main = async (
...
 }
 
-if (process.argv[1] === import.meta.filename) {
+if (
+  /* c8 ignore next */
+  (process.env.__VLT_INTERNAL_MAIN ?? process.argv[1]) ===
+  __CODE_SPLIT_SCRIPT_NAME
+) {
   process.title = 'vlt-cache-unzip'
```
isaacs, line 156: consider:

```ts
/* c8 ignore next */
const file = process.env.__VLT_INTERNAL_MAIN ?? process.argv[1]
if (file === __CODE_SPLIT_NAME) {
```

isaacs: Even though it's an immediate throwaway var, it makes the "what is this noise" clearer, and ends up being fewer lines to look at.

isaacs (approved):
Approved once tests pass and a few style nits. Land asap, please, so I can use the same approach in the new @vltpkg/cache-revalidate module.

infra/build/src/bins/vlix.ts
```diff
@@ -1,3 +1,7 @@
-process.argv.splice(2, 0, 'install-exec')
-import vlt from '@vltpkg/cli-sdk'
-await vlt()
+if (process.env.__VLT_INTERNAL_MAIN) {
```
isaacs, line 1: add this:

```ts
delete process.env.__VLT_INTERNAL_MAIN
```

Same for other bins, too.

PR #555, opened 2025-03-20
rename src/vlt to src/cli-sdk, as pattern requires

isaacs:


lukekarrys (approved):

infra/build/src/bundle.ts
```diff
@@ -277,7 +277,7 @@ export const bundle = async ({
     join(outdir, define.REGISTRY_CLIENT_PACKAGE_JSON),
   )
   cpSync(
-    join(WORKSPACE_DIR, 'vlt/package.json'),
+    join(CLI, 'package.json'),
```
isaacs, line 280: Ahh, tricky tricky.

lukekarrys: I changed it to use the `CLI` variable instead of just changing `vlt/` to `cli-sdk/` so that NEXT time we move this workspace it won't be a problem 😅

PR #557, opened 2025-03-20
security-archive: fix cache ttl check

ruyadorno:
The value returned by the sqlite time method is still in seconds even when you add subsecond precision, this change will make sure it's normalized to ms so that the ttl check works as intended.


isaacs (approved):
Can be in a subsequent patch, but I think this shows we ought to have tests for the sql here, even if it's just snapshots that get manually checked and then never change.

PR #570, opened 2025-03-22
cli-sdk: add generic error stderr print

ruyadorno:
Error objects that have no `cause` property should still print something to `stderr` to acknowledge the error to the end user.

More specifically this fixes problems with validation code in the `@vltpkg/query` module that will throw errors that don't necessarily have an underlying `Error` cause but rather intentional validation performed by our implementation.

A simple example of usage with this PR is:

```
vlt query ':bork'
Error: Unsupported pseudo-class: :bork
```


lukekarrys:
Closing in favor of #572

PR #576, opened 2025-03-24
fix `rollback-remove` when compiled

lukekarrys:
This does for `rollback-remove` what #553 and #564 and did for `cache-unzip`.

Now that this pattern is better established it also adds some tests for it in both `cache-unzip` and `rollback-remove`.

It also adds a smoke test to assert that `rollback-remove` runs successfully when compiled by installing two different versions of the same package by-to-back and testing that the old version has no `.VLT.DELETE.` files left behind.


isaacs (approved):
Lgtm

PR #579, opened 2025-03-26
error-cause: more generic type checks

lukekarrys:
This removes the `isErrorRoot` type guard in favor of a new one `isErrorWithCode`. This better matches what is actually being done with the type guard in `print-err`.

Previously it was only checking if the error had a cause that was an object, but asserting that the cause was one thrown by `@vltpkg/error-cause`. Now it is checking if the code is one of our known codes.

This also takes any `cause.cause` (which has to be `unknown` for compatability with TS) and converts it to an `Error` and updates the resulting types appropriately.


ruyadorno (approved):
LGTM 👍

PR #580, opened 2025-03-26
stale-while-revalidate

isaacs:
This adds `stale-while-revalidate` support to the registry-client.

### DIFF
--- a/infra/smoke-test/test/cache-unzip.ts
+++ b/infra/smoke-test/test/cache-unzip.ts
@@ -1,10 +1,10 @@
+import { CacheEntry } from '@vltpkg/registry-client/cache-entry'
+import { readdirSync, readFileSync } from 'node:fs'
+import { join } from 'node:path'
+import { setTimeout } from 'node:timers/promises'
 import t from 'tap'
 import { runMultiple } from './fixtures/run.ts'
 import { defaultVariants } from './fixtures/variants.ts'
-import { setTimeout } from 'node:timers/promises'
-import { readdirSync, readFileSync } from 'node:fs'
-import { join } from 'node:path'
-import { CacheEntry } from '@vltpkg/registry-client/cache-entry'
 
 t.test(
   'unzips all cache entries after a successful install',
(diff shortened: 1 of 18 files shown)


src/registry-client/src/cache-revalidate.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+    proc.stdin.end()
+    // Another Deno oddity. Calling unref on a spawned process will kill the
+    // process unless it is detached. https://github.com/denoland/deno/issues/21446
+    // So in this case Deno on Windows will be slower to exit the main process
+    // since it will wait for the child process to exit.
+    // TODO: figure out something better to do here?
+    if (detached) {
+      proc.unref()
+    }
```
lukekarrys, line 75: If we write this block a 4th time, I think it should be abstracted to its own workspace. The end goal is to not need to do this because we'll settle on a runtime and testing strategy. But then we can update and test the logic in only a single place.

src/registry-client/src/index.ts
```diff
@@ -133,12 +164,14 @@ const { version } = loadPackageJson(
 ) as {
   version: string
 }
+/* c8 ignore start - we do test this, but coverage fails */
```
lukekarrys, line 167: This is probably because https://github.com/vltpkg/vltpkg/blob/main/CONTRIBUTING.md#test-coverage-is-failing-but-it-shouldnt-be.

Something with using type stripping messes with coverage but only when doing `import from X` and `t.mockImport(X)` in the same test.

infra/smoke-test/test/cache-unzip.ts
```diff
@@ -16,9 +16,12 @@ t.test(
         // time but should be enough for a small install.
         await setTimeout(1000)
 
-        const cacheEntries = readdirSync(join(dirs.cache, 'vlt'), {
-          withFileTypes: true,
-        })
+        const cacheEntries = readdirSync(
+          join(dirs.cache, 'vlt/registry-client'),
```
lukekarrys, line 20: Does this PR change the location of our cache directory?

lukekarrys (approved):
Approved with one question about the cache-unzip smoke test to help me understand if I was testing the right thing.

Failing smoke test is a flake that I will continue to debug elsewhere.

infra/smoke-test/test/cache-unzip.ts
isaacs: Previously, the config would override the xdg location, and would default to having 'registry-client' in the path. But if configured, it would not. Now, the default is just xdg.cache(), and registry-client is tacked on regardless of whether it's the default or the configured value.

PR #601, opened 2025-04-02
add 'vlt cache' command

isaacs:
```
Usage:
  vlt cache <command> [flags]

Work with vlt cache folders

  Subcommands

    add
      Resolve the referenced package identifiers and ensure they are cached

      ​vlt cache add <package-spec> [<package-spec>...]

    ls
      Show cache entries

      ​vlt cache ls [<key>...]

    clean
      Purge expired cache entries

      ​vlt cache clean [<key>...]

    delete
      Purge items explicitly, whether expired or not

      ​vlt cache delete <key> [<key>...]

    delete-before
      Purge all cache items from before a given date

      ​vlt cache delete-before <date>

    delete-all
      Delete the entire cache folder to make vlt slower

      ​vlt cache delete-all

```

### DIFF
--- a/src/cli-sdk/src/commands/cache.ts
+++ b/src/cli-sdk/src/commands/cache.ts
@@ -1,0 +1,312 @@
+import { error } from '@vltpkg/error-cause'
+import { CacheEntry } from '@vltpkg/registry-client'
+import { Spec } from '@vltpkg/spec'
+import { mkdir, rm } from 'node:fs/promises'
+import prettyBytes from 'pretty-bytes'
+import type { LoadedConfig } from '../config/index.ts'
+import type { CommandUsageDefinition } from '../config/usage.ts'
+import { commandUsage } from '../config/usage.ts'
+import type { CommandFn, CommandUsage } from '../index.ts'
+import { stdout } from '../output.ts'
+import type { ViewOptions, Views } from '../view.ts'
+import { ViewClass } from '../view.ts'
+
+export type CacheMap = Record<
+  string,
+  ReturnType<CacheEntry['toJSON']>
+>
+
+export type CacheSubcommands = keyof (typeof usageDef)['subcommands']
+
+let view: CacheView
+export class CacheView extends ViewClass {
+  constructor(options: ViewOptions, conf: LoadedConfig) {
+    super(options, conf)
+    view = this
+  }
+  stdout(...args: unknown[]) {
+    stdout(...args)
+  }
+  error(e: unknown) {
+    throw e
+  }
+}
+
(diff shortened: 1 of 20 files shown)


src/cli-sdk/src/commands/cache.ts
```diff
@@ -0,0 +1,306 @@
...
+        view?.stdout(
+          key.includes(' ') ? JSON.stringify(key) : key,
+          entry,
+        )
+        return true
+      },
+      view,
+    )
+  : await fetchAll(conf, (_, key) => {
+      //view?.stdout(key)
```
ruyadorno, line 131: oops?

src/cli-sdk/src/commands/cache.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  subcommands: {
+    add: {
+      usage: '<package-spec> [<package-spec>...]',
+      description:
+        'Resolve the referenced package identifiers and ensure they are cached',
+    },
+
+    ls: {
+      usage: '[<key>...]',
```
ruyadorno, line 52: what are `<key>` in this context? integrity values?

src/cli-sdk/src/commands/cache.ts
```diff
@@ -0,0 +1,306 @@
...
+  return map
+}
+
+const fetchKeys = async (
+  conf: LoadedConfig,
+  keys: string[],
+  test: (entry: CacheEntry, key: string, buf: Buffer) => boolean,
+  view?: CacheView,
+) => {
+  const rc = new RegistryClient(conf.options)
```
ruyadorno, line 157: is there a `RegistryClient` instance in `conf` that can be used here instead of the new instance? not a necessary change but I wonder if there is a reusable instance already.

ruyadorno (commented):
It might be useful to add usage examples for `vlt cache ls <key>` and `vlt cache delete-before` - as an user I appreciate these kind of references on how to use the parameter / values.

ruyadorno:
other than that it LGTM 👍

src/cli-sdk/src/commands/cache.ts
isaacs: No, the keys that are printed out if you do `vlt cache ls` without any arguments. (Most of time, these will be URLs.)

src/cli-sdk/src/commands/cache.ts
isaacs: No, we don't create a RegistryClient, though there is one attached to the packageInfo client that's already there. I could use that 👍

Issue #592, opened 2025-03-31
[FEATURE] Add `--expect-results=[<boolean>|<numeric>]` to `vlt query`


darcyclarke:
### Description
We should be able to give `query` the ability to error when the result of a query is not what was expected (ie. either by a numeric value - ex. `1`, `>=`, `0`, `<100` - or by a boolean - ex. `false`/`true`)

PR #604, opened 2025-04-03
cli-sdk: add expect-results config

ruyadorno:
Add a new `--expect-results` config value that when used with the `vlt query` command allows for validation of the result values of a query.

Example:

    vlt query --expect-results=">=1"
    my-project
    └── a@1.0.0

    vlt query --expect-results=false
    Error: Unexpected number of items
        at ...
      [cause]: { found: 1, wanted: 0 }
    }

Fixes: https://github.com/vltpkg/vltpkg/issues/592

### DIFF
--- a/src/cli-sdk/src/commands/query.ts
+++ b/src/cli-sdk/src/commands/query.ts
@@ -25,7 +25,7 @@ export const usage: CommandUsage = () =>
     command: 'query',
     usage: [
       '',
-      '<query> --view=[human | json | mermaid | gui]',
+      '<query> --view=<human | json | mermaid | gui>',
       '<query> --expect-results=<comparison string>',
     ],
     description:
--- a/src/cli-sdk/src/commands/query.ts
+++ b/src/cli-sdk/src/commands/query.ts
@@ -1,4 +1,5 @@
 import type {
+  EdgeLike,
   HumanReadableOutputGraph,
   JSONOutputGraph,
   MermaidOutputGraph,
@@ -10,17 +11,23 @@ import {
   jsonOutput,
   mermaidOutput,
 } from '@vltpkg/graph'
+import { error } from '@vltpkg/error-cause'
 import { Query } from '@vltpkg/query'
 import { SecurityArchive } from '@vltpkg/security-archive'
 import { commandUsage } from '../config/usage.ts'
 import type { CommandFn, CommandUsage } from '../index.ts'
 import { startGUI } from '../start-gui.ts'
 import type { Views } from '../view.ts'
+import type { LoadedConfig } from '../config/index.js'
 
 export const usage: CommandUsage = () =>
   commandUsage({
     command: 'query',
-    usage: ['', '<query> --view=[human | json | mermaid | gui]'],
+    usage: [
+      '',
+      '<query> --view=[human | json | mermaid | gui]',
+      '<query> --expected-results=[number | boolean | string]',
+    ],
     description:
       'List installed dependencies matching the provided query.',
     examples: {
@@ -38,8 +45,16 @@ export const usage: CommandUsage = () =>
         description:
           'Query packages with names starting with "@vltpkg"',
       },
+      [`'*:license(copyleft) --expect-results=0'`]: {
+        description: 'Errors if a copyleft licensed package is found',
+      },
     },
     options: {
+      'expect-results': {
+        value: '[number | boolean | string]',
+        description:
+          'Sets an expected number of resulting items. Errors if the number of resulting items does not match the set value. Accepts a specific numeric value, "true" (same as "> 0"), "false" (same as 0) or a string value starting with either ">", "<", ">=" or "<=" followed by a numeric value to be compared.',
+      },
       view: {
         value: '[human | json | mermaid | gui]',
         description:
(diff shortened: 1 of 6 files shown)


src/cli-sdk/src/commands/query.ts
```diff
@@ -52,6 +67,29 @@ type QueryResult = JSONOutputGraph &
   MermaidOutputGraph &
   HumanReadableOutputGraph & { queryString: string }
 
+const validateExpectedResult = (
```
isaacs, line 70: This validation function belongs in `src/cli-sdk/src/config/definitions.ts`, so that incorrect configs will be validated in the same way as all other configurations.

isaacs: Oh, no, this is a different thing, nevermind.

src/cli-sdk/src/config/definition.ts
```diff
@@ -545,3 +545,19 @@ export const definition = j
...
+                    set a expected number of resulting items.
+
+                    Accepted values are numbers, booleans: true (to expect at
+                    least one result), false (to expect no results) and
+                    strings.
+
+                    Strings starting with \`>\`, \`<\`, \`>=\` or \`<=\`
+                    followed by a number can be used to check if the result
+                    is greater than or less than a specific number.`,
+    },
```
isaacs, line 562: Mixing boolean and string types is an antipattern. It would be the only boolean that is not a flag, gets set to `'true'` in the environment instead of `1` for true and `0` for false, etc.

It would be better to use `--expect-results=">0"` instead of `"true"`, which means the same thing, or `--expect-results=0` instead of `false`.

This definition should have a validation function, like:

```ts
validate: (value: string) => /^([<>]=?)?[0-9]+$/.test(value)
```

Then we can be ensured that no matter where the value comes from, it'll match the expected value.

src/cli-sdk/src/commands/query.ts
```diff
@@ -52,6 +67,29 @@ type QueryResult = JSONOutputGraph &
...
+  const expectResults = conf.values['expect-results']
+  if (expectResults?.startsWith('>=')) {
+    return edges.length >= parseInt(expectResults.slice(2).trim(), 10)
+  } else if (expectResults?.startsWith('<=')) {
+    return edges.length <= parseInt(expectResults.slice(2).trim(), 10)
+  } else if (expectResults?.startsWith('>')) {
+    return edges.length > parseInt(expectResults.slice(1).trim(), 10)
+  } else if (expectResults?.startsWith('<')) {
+    return edges.length < parseInt(expectResults.slice(1).trim(), 10)
+  } else if (expectResults === 'true') {
```
isaacs, line 83: We should not have pseudo-boolean non-flags in our config set.

src/cli-sdk/src/commands/query.ts
```diff
@@ -10,17 +11,23 @@ import {
...
+import type { LoadedConfig } from '../config/index.js'
 
 export const usage: CommandUsage = () =>
   commandUsage({
     command: 'query',
-    usage: ['', '<query> --view=[human | json | mermaid | gui]'],
+    usage: [
+      '',
+      '<query> --view=[human | json | mermaid | gui]',
+      '<query> --expect=[number | boolean | string]',
```
isaacs, line 29:
```suggestion
      '<query> --expect-results=[number | boolean | string]',
```

isaacs (changes requested):
Get rid of `'true'` and `'false'` as values. Just use `--expect-results=">0"` and `--expect-results=0`. Also, typo in the command usage text.

darcyclarke:
I agree with @isaacs feedback that we should just use the numeric values although I'd add that it'd be ideal to set a default value of `">0"` when `--expect-results` is passed but no value was specified. That change would essentially recreate the truthiness of the flag & is inline with its name.

src/cli-sdk/src/commands/query.ts
```diff
@@ -10,17 +11,23 @@ import {
...
+import type { LoadedConfig } from '../config/index.js'
 
 export const usage: CommandUsage = () =>
   commandUsage({
     command: 'query',
-    usage: ['', '<query> --view=[human | json | mermaid | gui]'],
+    usage: [
+      '',
+      '<query> --view=[human | json | mermaid | gui]',
+      '<query> --expect-results=[number | string]',
```
isaacs, line 29:
```suggestion
      '<query> --expect-results=<comparison string>',
```

Or something to clarify it's a specific _sort_ of string we're after here.

src/cli-sdk/src/commands/query.ts
```diff
@@ -10,17 +11,23 @@ import {
...
 import type { Views } from '../view.ts'
+import type { LoadedConfig } from '../config/index.js'
 
 export const usage: CommandUsage = () =>
   commandUsage({
     command: 'query',
-    usage: ['', '<query> --view=[human | json | mermaid | gui]'],
+    usage: [
+      '',
+      '<query> --view=[human | json | mermaid | gui]',
```
isaacs, line 28:
```suggestion
      '<query> --view=<human | json | mermaid | gui>',
```

Technically, the `[]` should be for optional components of an example argument, and `<>` for required ones, and `--view=` (no value) isn't valid.

isaacs (approved):
Some minor suggestions, but this is fine.

Another idea, perhaps: instead of `--expect-results=">0"` (since `<>` need to be quoted on a CLI), what if we did `--min-results=<number>` and `--max-results=<number>`?

Then they can both be just number types in Jackspeak, which already has built-in validation and such, and the test is much simpler. Just make the default `min-results` be 0, and the default `max-results` be Infinity, and the test is always `count <= max && count >= min`, no need for special characters or anything. To say "no results" you set `--max-results=0`, to say "some results" you set `--min-results=1`.

I guess then if you expect some _specific number_ of results, it's somewhat harder, because you have to set them both the same value, but it seems quite unusual to want specifically 7 results of a query for some reason.

PR #605, opened 2025-04-03
query: fix query response types

ruyadorno:
Adds a `QueryResponseEdge` type so that both the `nodes` and `edges` returned from `query.search()` can have a unified type definition.


ejkorol (approved):
very easy 👍

PR #606, opened 2025-04-03
graph: fix aliases in human readable output

ruyadorno:
Fix the representation of aliased dependencies and custom registry prefixes in the human readable output.

The registry prefix now goes along with the node name/version instead of prefixing the dependency name.

## Example

For a given `package.json` file:

```json
{
  "name": "my-project",
  "dependencies": {
    "a": "npm:abbrev@^3.0.0"
  }
}
```

### Before:

```
    vlt ls
    my-project
    └── npm:a (abbrev@3.0.0)
```

### After:

```
    vlt ls
    my-project
    └── a (npm:abbrev@3.0.0)
```


isaacs (approved):
LGTM!

isaacs:
Ah, looks like you need to regenerate snapshots in `src/cli-sdk` and then it'll pass.

PR #612, opened 2025-04-04
query: add id selector scoped names and semver

ruyadorno:
Add support to scoped names and semver range definitions to the id selector. In order to work around invalid characters, escaping is required, e.g:

    vlt query "#minimatch@9\.0\.0"

The complete semver dialect should be available:

    vlt query "#minimatch@9"
    vlt query "#minimatch@^9\.0\.0"
    vlt query "#minimatch@\>=8"
    vlt query "#minimatch@\3\ \|\|\ 9"

Scoped names require escaping the slash, e.g:

    vlt query "#@babel\/core"


darcyclarke:
Oof, the escaping is definitely not clean & seems pretty inconsistent/annoying as an end-user. 

I know what you're going for but I can't see anyone writing these escaped versions of selectors by hand without running into issues (ex. I just tried using this programmatically & ran into the problem where I had to double escape the selector in the string - ie. `#@babel\/core` became `"#@babel\\/core"` - really annoying). What's happening here is that we're leaking the selector parser's inability to handle these values because they don't align with CSS. We can't say DSS is a superset of CSS if we're going to add invalid syntax like `#@` so we might as well make exceptions for these, very common, shorthands.

I think we have two options if you agree: 
1. consider modifying that parser to support these as first-class shorthands (basically means we'd fork it which isn't ideal - unless they have some handy way of changing the token logic)
2. do preprocessing like we did with `npm`'s implementation - this means that the underlaying parser never actually sees these short-hand selectors, it sees `[name=<name>]` & `[name=<name>]:semver(<version>)` (as those should be functionally equivalent to the shorthands)

### Currently:
```
#minimatch
#minimatch@8
#minimatch@\>=8
#@babel\/core
#@babel\/core@9
#@babel\/core@^9\.0\.1
```
### Expected:
```
#minimatch
#minimatch@8
#minimatch@>=8
#@babel/core
#@babel/core@9
#@babel/core@^9.0.1
```

darcyclarke (changes requested):
See: https://github.com/vltpkg/vltpkg/pull/612#issuecomment-2781166615

ruyadorno:
> ex. #foo@1.x.prod is x a class or part of the version?

For me that's the parsing logical problem that made me make peace with the otherwise very aesthetically unappealing escaping of characters.

The other strong argument in favour of escaping is that one of the very likely possibilities in the future of the tooling around query is that we end up with more queries being generated by UI than hand-written by the end-user. Today we have the point-and-click navigation that generates queries but talking with @ejkorol this week he also have ideas on specialized UIs to guide the user on query-creation, that would also very much alleviate the problem. A worth noticing, third, UI-generated query usage is the point-and-click interface to generate overrides.

Given the impossibility of making semver versions work without syntax confusion and my personal view that escaping is not so unpractical of a solution. I'm -1 on trying to make id selectors semver work without the escaping proposed in this PR.

darcyclarke:
I understand we'll be able to provide a query builder/UI in the future but if that exists then the underlaying syntax isn't as important for that end-user. Where it is important to consider is for anyone that would write these by hand either in the programmatic or CLI usages.

I personally think we should follow npm here somewhat & drop the version. If you want to select by version us `[version=...]` or `:semver(...)`. Also, npm supports unescaped scoped names as ids right now so it feels like a downgrade/regression for us to deviate here.

So the updated set of selectors would be:

```
#minimatch
#minimatch:semver(8)
#minimatch:semver(>=8)
#@babel/core
#@babel/core:semver(9)
#@babel/core:semver(^9.0.1)
```

If we're considered about the look/feel/size of these selectors & would prefer a more minimalist API for versions/semver ranges we could consider adding a pseudo-class alias (ex. `:v(...)`). This would make it a bit simpler, ex. `#minimatch:semver(>=8)` becomes `#minimatch:v(>=8)` & `#@babel/core:semver(^9.0.1)` becomes `#@babel/core:v(^9.0.1)`.

PR #618, opened 2025-04-07
query socket score

ejkorol:
- **adds the socket `score?` to the insights**
- **upates query tests to contain the socket score property**


ruyadorno (changes requested):
Let's nest it inside the `insights` object 🙏

src/query/src/index.ts
```diff
@@ -204,6 +204,9 @@ export class Query {
       // if a security archive entry is found then we can populate the insights
       node.insights = {
         scanned: true,
+        score: {
+          ...securityArchiveEntry.score,
+        },
```
ruyadorno, line 209:
```suggestion
        score: securityArchiveEntry.score,
```

ruyadorno (approved):

PR #621, opened 2025-04-08
cache-unzip: properly read length/encoding headers

lukekarrys:
When unzipping cache entries we use the modulo operator to know when we are reading a header key or value.

Previously this operation was accidentally being performed on the offset which was only passing tests by coincidence that the keys and values were even lengths.

This commit makes sure we are checking the length of the headers array instead and adds some oddly length-ed headers to the test.


isaacs (approved):
Ha! Looking at the headers that the registry sends, it works by accident there, too.

Good catch!

PR #629, opened 2025-04-09
gui: socket score

ejkorol:
- **adds tremor ui `progress` components for package scores**
- **chore: removes unused image asset**
- **bugfix: uses window location for history**
- **creates general purpose link component**
- **adds `packageScore` to `selected-item` context**
- **removes `insights` component from `selected-item`**
- **adds `packageScores` into `tabs-insights`**
- **- integrates the `insights` into `item-header`**


src/gui/src/components/explorer-grid/selected-item/item-header.tsx
```diff
@@ -126,122 +133,199 @@ const Downloads = () => {
...
+    </div>
+  )
+}
+
+const PackageOverallScore = () => {
+  const { packageScore, setActiveTab } = useSelectedItem()
+
+  if (!packageScore) return null
+
+  const overallScore = packageScore.overall * 100
```
ruyadorno, line 193: I saw some places in which this operation did not turned out well, let's `Math.round` here too be on the safe side.

```suggestion
  const overallScore = Math.round(packageScore.overall * 100)
```

### Reference: 

Screenshot 2025-04-10 at 9 24 16 AM

src/gui/src/components/explorer-grid/selected-item/tabs-insight.tsx
```diff
@@ -113,6 +116,60 @@ const InsightItem = ({
...
+    'maintenance',
+    'vulnerability',
+    'license',
+  ]
+
+  const scores = Object.entries(packageScore)
+    .map(([key, value]) => ({
+      id: key,
+      name: key === 'supplyChain' ? 'supply chain' : key,
+      value: value * 100,
```
ruyadorno, line 137: Same as the other related comment, let's err on the safe side here.

```suggestion
      value: Math.round(value * 100),
```

ruyadorno (changes requested):

PR #633, opened 2025-04-09
cache: handle all integrity cases when writing cache entry to disk

### LINKED ISSUE #627: [BUG] `cache.set` with integrity should work for all use cases
Discussed with @isaacs. Here's the notes from our Slack convo.

This stems from #622 which fixed the bug where integrity files were not being ungzipped. But the fix was a bit of a hack and a more robust solution that lives in `@vltpkg/cache` is necessary.

It's important to consider the case where two packages from different registries can have the same integrity but different cache keys, which is common when mirroring or upstreaming (as `vsr` is planned to support).

So `cache.write(key, val, { integrity })` has to handle the following cases:

1. integrity file isn't present
    - This is what happens when you first write any entry to the cache
    - **Action:** write the value file, hardlink to integrity path
1. integrity file is present, links to value file
    - This will happen when you cache-unzip the first package
    - **Action:** write the value file, hardlink to integrity path (overwriting it, break all other links)
1. integrity file is present, does not link to value file, but contents match
    - This will happen when you try to do the second write (new key, matching integrity), so now you have nlink=3 on that file
    - **Action:** just hardlink to integrity file
1. integrity file is present, does not link to value file, contents don't match
    - This will happen if somehow the link is broken (eg, if cache-unzip fails halfway through or something else weird happens)
    - **Action:** write value file, hardlink to integrity path (overwriting it, break all other links)

`cache-unzip` should also `await cache.promise()` after each `cache.set()` due to the potential race condition between two writes with the same integrity. This is slower, but happens in a separate process so its ok.

lukekarrys:
Also update cache-unzip to use the reworked `cache.set` method with the
`integrity` option.

Fixes #627

### DIFF
--- a/src/cache/src/index.ts
+++ b/src/cache/src/index.ts
@@ -55,10 +55,11 @@ export type BooleanOrVoid = boolean | void
 const hash = (s: string) =>
   createHash('sha512').update(s).digest('hex')
 
-const success = <T>(p: Promise<T>): Promise<T | null> =>
+const FAILURE = null
+const success = <T>(p: Promise<T>): Promise<T | typeof FAILURE> =>
   p.then(
     result => result,
-    () => null,
+    () => FAILURE,
   )
 
 export class Cache extends LRUCache<
@@ -413,7 +414,7 @@ export class Cache extends LRUCache<
       // Path doesn't exist but integrity is already what we want
       // Try to link integrity file to val file
       // If linking fails, continue on
-      if ((await success(link(intFile, valFile))) !== null) {
+      if ((await success(link(intFile, valFile))) !== FAILURE) {
         await writeKeyP
         return
       }
@@ -434,7 +435,8 @@ export class Cache extends LRUCache<
       // so its likely the same content so force link them
       // because thats faster than writing the value file.
       if (
-        (await success(this.#linkAtomic(intFile, valFile))) !== null
+        (await success(this.#linkAtomic(intFile, valFile))) !==
+        FAILURE
       ) {
         await writeKeyP
         return
--- a/src/cache/src/index.ts
+++ b/src/cache/src/index.ts
@@ -440,7 +440,7 @@ export class Cache extends LRUCache<
 
     // If files are not same size or something went wrong with the previous link
     // operations then fallback to doing everything:
-    // - write the write the key/value files
+    // - write the key/value files
     // - unlink integrity and then re-link it to the value file
     await Promise.all([
       writeKeyP,
--- a/src/cache/src/index.ts
+++ b/src/cache/src/index.ts
@@ -56,6 +56,7 @@ const hash = (s: string) =>
   createHash('sha512').update(s).digest('hex')
 
 const FAILURE = null
+
 const success = <T>(p: Promise<T>): Promise<T | typeof FAILURE> =>
   p.then(
     result => result,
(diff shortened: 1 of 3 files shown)


lukekarrys:
Still needs to tests but the `cache-unzip` and `smoke-test` are passing locally which means it's at least working for the happy path.

src/cache/src/index.ts
```diff
@@ -347,36 +355,99 @@ export class Cache extends LRUCache<
...
+    }
+
+    if (!linked && sameSize) {
+      // Integrity and val file are not linked but same size
+      // so its likely the same content so force link them
+      // because thats faster than writing the value file.
+      if (
+        (await success(
+          unlink(valFile).then(() => link(intFile, valFile)),
+        )) !== null
```
lukekarrys, line 434: @isaacs Calling out this block specifically because it should be the only difference from when we went over this in pairing.

Now if the value and integrity files are not linked, but are likely the same, it tries to force them to be linked. 

This should (almost?) always be what we want. The alternative of only calling `link(intFile, valFile)` without first unlinking will always reject with `EEXIST` which renders this block meaningless.

lukekarrys:
@isaacs This is ready for review!

src/cache/src/index.ts
```diff
@@ -54,28 +56,25 @@ export type BooleanOrVoid = boolean | void
 const hash = (s: string) =>
   createHash('sha512').update(s).digest('hex')
 
+const success = <T>(p: Promise<T>): Promise<T | null> =>
+  p.then(
+    result => result,
+    () => null,
```
isaacs, line 62: Kind of a style suggestion, but it might be good to have something like:

```ts
const FAILURE = null
const success = <T>(p: Promise<T>): Promise<T | typeof FAILURE> =>
  p.then(
    result => result,
   () => FAILURE,
  )
```

Then, later on, you could do `if (FAILURE === success(...))` and it's slightly more obvious why `null` is being used there.

src/cache/src/index.ts
```diff
@@ -347,36 +347,107 @@ export class Cache extends LRUCache<
       return this.integrityPath(i)
     } catch {}
   }
+
+  async #writeFileAtomic(file: string, data: Buffer | string) {
+    // ensure we get a different random key for every write,
+    // just in case the same file tries to write multiple times,
+    // it'll still be atomic.
+    const tmp = `${file}.${this.#random}${this.#i++}`
```
isaacs, line 355:
```suggestion
    const tmp = `${file}.${this.#random}.${this.#i++}`
```

isaacs: Not a significant functional difference, it's just as unique without the `.`, but that delimiter makes it clearer from the filenames that this is the nth write for a given Cache instance, which has been handy for debugging a few times.

src/cache/src/index.ts
isaacs: Correct, in this case you do know that you need to unlink the valFile, because you know it exists.

src/cache/src/index.ts
```diff
@@ -347,36 +347,107 @@ export class Cache extends LRUCache<
...
+        )) !== null
+      ) {
+        await writeKeyP
+        return
+      }
+    }
+
+    // If files are not same size or something went wrong with the previous link
+    // operations then fallback to doing everything:
+    // - write the write the key/value files
```
isaacs, line 443:
```suggestion
    // - write the key/value files
```

src/cache/src/index.ts
isaacs: Oh, actually, this should be an atomic link+rename, though, not an unlink and link, so that it's one fs op and doesn't ever have the file go missing.

Need something like this:

```ts
async #linkAtomic (src: string, dest: string) {
  const tmp = `${dest}.${this.#random}${this.#i++}`
  await link(src, tmp).then(() => rename(tmp, dest))
}
```

```suggestion
          this.#linkAtomic(intFile, valFile)),
        )) !== null
```

isaacs (changes requested):
One case that I'm realizing might be inefficiently handled here is the "not same size" scenario.

The most likely way that can happen is if we've already un-gzipped the integrity file.

For example, let's say we have `https://example.com/abbrev/-/abbrev-3.0.1.tgz` (A) and `https://registry.npmjs.org/abbrev/-/abbrev-3.0.1.tgz` (B), which are the same artifact at two different URLs.

- Fetch (A), and save keyFile, valFile, and integrityFile.
- Fetch (B), save keyFile, link valFile (because integrity/size match).

Then, some time later, we unzip.

- Write (A) with a larger body, and (correctly) clobber both the valFile and integrityFile, unlinking (B).
- _Most_ of the time, the subsequent write to (B) will be the same unzipped contents, and so it'll do the same integrity/size match re-link. But let's say that crashed and now some other process is trying to save the gzipped (B) contents back again. It's going to find the integrity file, see that it's not the same size, and clobber both.

Maybe the "sameSize" restriction should really be "size greater than or equal"? If it has the same integrity, we know it at least _started_ from the same bits, and then can only have been changed by virtue of being unzipped.

Also, "size" should really be "size of the body", not "size of the file", because a different registry (the only case being handled here) is going to almost certainly have different headers, but we should return the same response CacheEntry anyway if the integrity matches. This will work by accident if the (A) and (B) requests don't race with one another, but if they do, then each will clobber the "wrong size" integrity file each time they write (including for cache-unzip writes), which is inefficient.

This sounds like a sticky problem, but if we can trust our own integrity values, it actually might simplify it considerably.

- if !intFile, write valFile and be done (same as here)
- if !intStats, write valFile and link (same as here)
- if intStats and !valStats, link valFile to intFile (ignore sameSize test, we know it's the same bits)
- if linked and sameSize, just await key (same as here)
- if linked and !sameSize, atomic write intFile, then atomic link to valFile (the link empowers us to break any other users of this intFile)
- if !linked, atomic link intFile to valFile (presumably some other cache write has already handled or will be handling the content change, but we know it's functionally "the same" response body)

We should only use `this.#atomicLink` when we expect that we're clobbering a destination of the link; otherwise, the normal `link` is appropriate, because we will _want_ to fail with EEXIST, and the `link(1)` operation is already atomic, so it's really more like "clobbering link" than "atomic link", I guess. Idk, naming things is hard.

isaacs:
Oh, wait, there's still the "if not sameSize and linked, break the link and rewrite". Don't break the link unless you have the link.

Once that's done, all good.

src/cache/src/index.ts
```diff
@@ -376,81 +377,74 @@ export class Cache extends LRUCache<
...
-      if (
-        (await success(this.#linkAtomic(intFile, valFile))) !==
-        FAILURE
-      ) {
-        await writeKeyP
-        return
-      }
+    // We now know that both the value and integrity files exist.
+
+    // TODO: we are handling different registries here which will likely
```
isaacs, line 422: I think this TODO is kind of unnecessary, tbh, and it might not be worth the hassle.

If the size matches, that means that it was the same kind of thing, likely from the same exact URL. So, "do nothing" is completely correct. And if it's not, ok, well, we have the link, so breaking the link is allowed.

isaacs (approved):
Nice cleanup. Still LGTM.

PR #648, opened 2025-04-11
query: no longer support classes

ruyadorno:
Refactor the query language so that it no longer supports classes.

Dependency type filters are now transfered to pseudo selectors:

    .dev -------> :dev
    .optional --> :optional
    .peer ------> :peer
    .prod ------> :prod
    .workspace -> :workspace

This refactors also avoids issues parsing the parameters of pseudo selectors that use any dots such as semver and others, by escaping any found dot value e.g:

    :semver(2.0.0)
    :score(>0.5)
    :published(2024-01-01T11:11:11.111Z)


src/query/src/parser.ts
```diff
@@ -9,11 +11,24 @@ import type { Root } from 'postcss-selector-parser'
 export const escapeScopedNamesSlashes = (query: string): string =>
   query.replace(/(#@\w+)\//gm, (_, scope: string) => `${scope}\\/`)
 
+export const escapeDots = (query: string): string =>
+  query.replace(/\./gm, '\\.')
```
lukekarrys, line 15:
```suggestion
  query.replaceAll('.', '\\.')
```

src/query/src/parser.ts
```diff
@@ -9,11 +11,24 @@ import type { Root } from 'postcss-selector-parser'
...
+
+export const unescapeDots = (query: string): string =>
+  query.replace(/\\\./gm, '.')
```
lukekarrys, line 18:
```suggestion
  query.replaceAll('\\.', '.')
```

lukekarrys (approved):

Issue #357, opened 2025-01-28
[CHORE] GUI: Navigating to a dependency should check & use shorthands if possible


darcyclarke:
### Description
When navigating dependencies the query selector automatically assumes the longest-form dependency reference by name (ie. `[name=foo]` or `[name=@foo/bar]`); that said, their is an alternative shorthand which can be used for unscoped packages (shortening the overall length of many queries) which is prefixing `#` to the package name (ex. `#foo`).

### Suggestion
- [ ] when clicking/navigating to a dependency check first if we can use the shorthand version of the query (ie. `#lodash` vs. `[name=lodash]`)
  - note, this only works for unscoped packages but would be a significant win & cut down on the size/length of queries

### Example (old)

![Image](https://github.com/user-attachments/assets/6f245450-de40-42c0-a329-9513d8c17b29)

### Example (new)

![Image](https://github.com/user-attachments/assets/bf8d932e-3dff-42a9-9a4e-e54b56b2cda6)

PR #652, opened 2025-04-12
gui: change query navigation tokens

ruyadorno:
Change the query bar navigation to use id selectors and the `:v()` pseudo selector instead of attribute selectors.

Fixes: https://github.com/vltpkg/vltpkg/issues/357
Fixes: https://github.com/vltpkg/statusboard/issues/98


darcyclarke (changes requested):
Notably, we can drop the version entirely now that id selectors are using the edge name, ie. when they're found in a path they are unique (ex. `:root > #foo > #bar > #@baz/qux` cannot be anything but the dependencies/edges we resolved for each of `"foo"`, `"bar"` & `"@baz/qux"`). That change would further slim down the selectors & has the benefit of being more resilient to changes (ex. if we eventually add the ability to update or override nodes then a selector with a version is going to break - so queries that have explicit versions in the path become very brittle).

Long story long - let's drop versions all together for navigation (ie. no more `:v()` in generated path selectors)

ruyadorno:
oooohhh great point! I'll update it accordingly 👍

ruyadorno:
> no more `:v()` in generated path selectors

Note that in its actual shape we would still have package/node/manifest-based matching generated when you click on dependents/ancestors, we would need to brainstorm a way to change _that_ logic in order to truly have no more `:v()` in generated path selectors.

darcyclarke:
After discussing this, we've agreed to drop the `:v()` from generated path selectors & will create a unique path/"breadcrumb" from the `:root` &/or the relevant importer (ie. `#foo:project > #bar`). Stacking may be effected by this but should still, when clicked to unstack, navigate the user to the unique path/breadcrumb for each item/edge. There's likely a necessary change upstream to `query` to make this easier (ie. adding the path to the item/edge class would make it trivial to navigate directly to that path/breadcrumb & determine ancestry from that as well)

darcyclarke (approved):

PR #654, opened 2025-04-14
docs: move query selector references to docs only + new docs

darcyclarke:
### Description
- moved the query selector references solely to the docs (linking back from the query usage/pkg readme)
- added new dep id doc
- added new lockfile doc


www/docs/src/content/docs/cli/dep-ids.md
```diff
@@ -0,0 +1,200 @@
...
+
+### 5. Remote Dependencies
+
+Remote dependencies are packages installed from a URL.
+
+```json
+"remote·url"
+```
+
+Example: `"remote·https%3A%2F%2Fexample.com"`
```
isaacs, line 95: That is not how we encode `/` characters, btw, because various systems get upset about folder paths containing `%2F`, as it looks like security shenanigans.

To get around this, we use the `§` section character as a stand-in for spaces.

```
> d.joinDepIDTuple(['remote', 'https://example.com/foo.tgz'])
'remote·https%3A§§example.com§foo.tgz'
```

PR #656, opened 2025-04-15
gui: selected item metadata

ejkorol:
### Overview Tab
Screenshot 2025-04-16 at 12 22 54

### Versions Tab
Screenshot 2025-04-16 at 16 11 46

### DIFF
--- a/src/gui/src/components/explorer-grid/selected-item/item-header.tsx
+++ b/src/gui/src/components/explorer-grid/selected-item/item-header.tsx
@@ -37,6 +37,7 @@ import { isRecord } from '@/utils/typeguards.js'
 import { formatDistanceStrict } from 'date-fns'
 import { labelClassNamesMap } from '../label-helper.ts'
 import { CopyToClipboard } from '@/components/ui/copy-to-clipboard.jsx'
+import { formatDownloadSize } from '@/utils/format-download-size.js'
 
 const SpecOrigin = ({
   item,
@@ -285,7 +286,7 @@ const PackageMetadata = ({ className }: { className?: string }) => {
           variant="mono"
           className={cn(INLINE_CODE_STYLES, 'gap-1 px-2')}>
           <Download size={16} className="mb-0.5" />
-          {(unpackedSize / 1000).toFixed(0)} KB
+          {formatDownloadSize(unpackedSize)}
         </InlineCode>
       )}
       {tarballUrl && (
--- a/src/gui/src/components/explorer-grid/selected-item/item-header.tsx
+++ b/src/gui/src/components/explorer-grid/selected-item/item-header.tsx
@@ -12,10 +12,17 @@ import {
   TooltipProvider,
   TooltipTrigger,
 } from '@/components/ui/tooltip.jsx'
-import { Home, Info } from 'lucide-react'
+import {
+  Home,
+  Scale,
+  ArrowBigUpDash,
+  EyeOff,
+  Download,
+  Target,
+} from 'lucide-react'
 import { InlineCode } from '@/components/ui/inline-code.jsx'
 import { Badge } from '@/components/ui/badge.jsx'
-import { SparkAreaChart } from '@/components/ui/spark-chart.jsx'
+import { SparkBarChart } from '@/components/ui/spark-chart.jsx'
 import { transformToWeeklyDownloads } from '@/utils/transform-weekly-downloads.js'
 import { splitDepID } from '@vltpkg/dep-id/browser'
 import { Spec } from '@vltpkg/spec/browser'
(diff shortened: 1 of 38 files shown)


src/gui/src/components/ui/copy-to-clipboard.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [copied, setCopied] = React.useState<boolean>(false)
+
+  const handleCopy = async () => {
+    if (!copyValue) return
+    try {
+      await navigator.clipboard.writeText(copyValue)
+      setCopied(true)
+      setTimeout(() => setCopied(false), 1500)
+    } catch (err) {
+      console.error('Failed to copy!', err)
```
ruyadorno, line 38: Wild idea, and def most an idea to think about, not something to change right now but I'm thinking that it would be a good idea to provide a fallback / graceful degradation here in this case, I'm not sure if it's going to be practical everywhere layout-wise but it would be nice to print the value in the html in case the copy failed, otherwise the user is never retrieving, let's say, the tarball value from the "Tarball" button in case that for some reason the clipboard API throws.

src/gui/src/components/explorer-grid/selected-item/tabs-overview.tsx
```diff
@@ -15,14 +24,80 @@ export const OverviewTabButton = () => {
 }
 
 export const OverviewTabContent = () => {
-  const { selectedItemDetails, selectedItem } = useSelectedItem()
+  const { manifest } = useSelectedItem()
```
ruyadorno, line 27: I'd rather normalize homepage, funding, bugs and keyword values in dedicated helper methods so that the template here doesn't need so many conditionals. On the plus side these helper methods tend to be simple and fun to unit test.

```suggestion
  const { manifest } = useSelectedItem()
  const homepage: string = getHomepage(manifest)
  const funding: string[] = getFunding(manifest)
  const bugs: { url?: string, email?: string} = getBugs(manifest)
  const keywords: string[] = getKeywords(manifest)
```

These helpers should probably live in `@vltpkg/types` but anyways, not something that needs to be addressed in this PR. It can be a later refactor.

ejkorol: I like this a lot, we could definitely create a bunch of helpers later on

src/gui/src/components/explorer-grid/selected-item/item-header.tsx
```diff
@@ -108,84 +118,195 @@ const Downloads = () => {
...
+          <InlineCode
+            variant="mono"
+            className={cn(INLINE_CODE_STYLES, 'gap-1 px-2')}>
+            <Scale size={16} className="mb-0.5" />
+            {manifest.license}
+          </InlineCode>
+        )}
+      {manifest.type && (
+        <InlineCode variant="mono" className={cn(INLINE_CODE_STYLES)}>
+          {manifest.type}
```
ruyadorno, line 256: I'd prefer printing `ESM` or `CJS`, it's also shorter!

```suggestion
          {manifest.type === 'module' : 'ESM' : 'CJS'}
```

src/gui/src/components/explorer-grid/selected-item/item-header.tsx
```diff
@@ -108,84 +118,195 @@ const Downloads = () => {
...
+        Object.entries(manifest.engines).map(
+          ([engine, version], idx) => (
+            <InlineCode
+              variant="mono"
+              className={cn(
+                'px-2 pl-7',
+                INLINE_CODE_STYLES,
+                engine === 'npm' && 'pl-8',
+              )}
+              key={`${engine}-${version}-${idx}`}>
```
ruyadorno, line 270: we need to limit the width of this item as apparently the `version` value can grow to be very wide:

Screenshot 2025-04-15 at 3 58 52 PM

src/gui/src/components/explorer-grid/selected-item/tabs-overview.tsx
```diff
@@ -15,14 +24,80 @@ export const OverviewTabButton = () => {
...
+          <div className="flex flex-col gap-1">
+            <p className="inline-flex items-center gap-1.5 text-sm text-muted-foreground">
+              <HeartHandshake size={16} className="text-rose-400" />
+              <span>Funding</span>
+            </p>
+            {Array.isArray(manifest.funding) ?
+              manifest.funding.map((entry, index) => (
+                <span key={index}>
+                  {typeof entry === 'string' ? entry : entry.url}
+                </span>
```
ruyadorno, line 57: I think this one should also be a `<Link href=...` similar to the conditions below, I stumbled upon a package that uses it and it renders the wrong font and it can't be clicked: 

Screenshot 2025-04-15 at 4 05 51 PM

src/gui/src/components/explorer-grid/selected-item/item-header.tsx
```diff
@@ -108,84 +118,195 @@ const Downloads = () => {
...
+              <GlyphIcon
+                color="green"
+                icon={engine as keyof typeof ICONS}
+                size="lg"
+                className={cn(
+                  'absolute left-2',
+                  engine === 'npm' && 'top-[0.2rem]',
```
ruyadorno, line 277: I think this special "npm` condition here is to make centralize the icon vertically? I personally find it to be not very well centered atm, also I would expect npm to be red instead of green.

Screenshot 2025-04-15 at 4 08 30 PM

darcyclarke (approved):

ruyadorno (approved):

PR #659, opened 2025-04-15
windows smoke test fixes

lukekarrys:
08c61365a7f92472082987f358df7a6702e94c59 fixed a bug where edges in were not being deleted when their parent node was deleted. This caused the `rollback-remove` instance to attempt to rename the same path twice, which was causing `EPERM` errors on Windows the 2nd time.

I was able to reproduce this consistently by running (with no node_modules/cache/lockfile) `vlt i abbrev@2.0.0` followed by `vlt i abbrev@3.0.0`.

01e8a34776fc2714f749f33b10d01d14aab74dfc fixed this error by treating `EPERM` errors the same as `ENOENT` but this caused the temporary `.VLT.DELETE.*` rollback files to never be removed which was causing the smoke-test to fail.

This PR keeps the handling of EPERM errors, because without it we get a lot of flaky tests. But it avoids this specific cause of this EPERM error by using the map of pending rollbacks to ensure that the same path is not rolled back more than once.

---

This PR also includes a few commits that made this easier to debug:

- local `.ps1` scripts for easier running of our source on Windows
- A `SMOKE_TEST_VARIANTS` env var to filter which smoke test variants are run by default

### DIFF
--- a/infra/smoke-test/test/fixtures/variants.ts
+++ b/infra/smoke-test/test/fixtures/variants.ts
@@ -9,23 +9,31 @@ import t from 'tap'
 // this makes the tests run faster
 export const Bins = ['vlt'] as const
 
-export const allVariants = [
-  'source',
-  'denoSource',
-  'bundle',
-  'denoBundle',
-  'compile',
-] as const
+const ALL_VARIANTS = {
+  source: 'source',
+  denoSource: 'denoSource',
+  bundle: 'bundle',
+  denoBundle: 'denoBundle',
+  compile: 'compile',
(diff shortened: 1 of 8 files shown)


lukekarrys:
Verifying that the smoke tests are now passing by running them manually (with `gh workflow run smoke-test.yml --ref=$(git rev-parse --abbrev-ref HEAD)`): https://github.com/vltpkg/vltpkg/actions/runs/14481515939

isaacs (approved):
Obviously correct improvement. 👍

PR #680, opened 2025-04-18
docs: update starlight

lukekarrys:


ruyadorno:
@lukekarrys it looks like this PR needs rebasing

PR #682, opened 2025-04-21
gui: bugfix selected item rerendering

ejkorol:
## Bugfix: Rerender Performance on selectedItem View

Previously, we managed `selectedItemDetails` through React state and triggered data fetching via `useEffect`. This caused multiple performance issues:
- Each update to details created a new object, triggering unnecessary re-renders—even when only part of the data changed.
- The `useEffect` ran twice on mount, causing up to 3 redundant renders.
- Switching to React `context` to avoid prop drilling made the issue worse: since `context` updates trigger re-renders for all consuming components, any state change (even unrelated) caused the entire `selectedItem` view to re-render.

This led to visual glitches like images and data flashing as each detail was set sequentially.

## Fix

This PR introduces a scoped `Zustand` store within the `selectedItem`’s context. `Zustand` allows us to subscribe to specific slices of state, ensuring components only re-render when the data they actually use changes.

This solution gives us the best of both worlds:
- We maintain clean component composition without prop drilling.
- We gain fine-grained reactivity and improved performance by eliminating unnecessary renders.

While alternatives like [use-context-selector](https://www.npmjs.com/package/use-context-selector) could achieve similar results, `Zustand` provides this behavior out-of-the-box with simpler ergonomics and aligns with the rest of our state management approach.

## Features

- Moving the `useEffect` that handled the `fetchDetails` into higher level within the `selectedItem` `context`
- Creating a `selectedItemStore` within state in `context`
- Creating a custom `useSelectedItemStore` function in `context` to access properties via a `selector`
- Updates all components that consume `selectedItem` to use the new store
- Updates tests and snapshots


src/gui/src/components/explorer-grid/selected-item/item-header.tsx
```diff
@@ -467,7 +487,9 @@ const PackageImageSpec = ({ className }: { className?: string }) => {
 }
 
 const Publisher = ({ className }: { className?: string }) => {
-  const { selectedItemDetails } = useSelectedItem()
+  const selectedItemDetails = useSelectedItemStore(
+    state => state.selectedItemDetails,
+  )
```
ruyadorno, line 492: don't we need to split this into multiple context retrieval in order to get more fine-grained control over the rerender? e.g:

1. Yank the avatar image render into its own component, that then gets the property as its own, single props, e.g:

```js
const HeaderAvatar = ({ publisherAvatar }) => (
  {publisherAvatar?.src && (
        <Avatar>
          <AvatarImage
            className="size-5 rounded-sm outline outline-[1px] outline-border"
            src={selectedItemDetails.publisherAvatar.src}
            alt={selectedItemDetails.publisherAvatar.alt}
          />
          <AvatarFallback className="flex size-5 items-center justify-center rounded-sm bg-secondary bg-gradient-to-t from-neutral-100 to-neutral-400 p-0.5 outline outline-[1px] outline-border dark:from-neutral-500 dark:to-neutral-800" />
        </Avatar>
  ) || ''}
)
```

2. Retrieve `publisherAvatar`:

```js
  const publisherAvatar = useSelectedItemStore(
    state => state.selectedItemDetails.publisherAvatar,
  )
```

3. Combine them in the result:

```js
return (
    <div
      className={cn(
        'flex h-[31.508px] items-center gap-2',
        className,
)}>
      <HeaderAvatar { ...{ publisherAvatar } } />
      ...
```

src/gui/src/components/explorer-grid/selected-item/context.tsx
```diff
@@ -48,59 +54,103 @@ const getSocketInsights = (
...
 }
 
-export interface SelectedItemContextValue {
+type SelectedItemStoreState = {
   selectedItem: GridItemData
-  selectedItemDetails: DetailsInfo
   activeTab: Tab
-  setActiveTab: (tab: Tab) => void
-  insights: SocketSecurityDetails[] | undefined
+  selectedItemDetails: DetailsInfo
```
ruyadorno, line 60: I would prefer keeping everything to a flat, single level instead of nesting the resulting properties of the `fetchDetails` call to this `selectedItemDetails`object for two reasons: not only it sticks better to the zustand model already in place but it also reduces the chance of accidentally incurring rerenders in case a component uses the `selectedItemDetails` object as a prop, wdyt?

src/gui/src/components/explorer-grid/selected-item/context.tsx
```diff
@@ -48,59 +54,103 @@ const getSocketInsights = (
     .filter(Boolean) as SocketSecurityDetails[]
 }
 
-export interface SelectedItemContextValue {
+type SelectedItemStoreState = {
```
ruyadorno, line 57:
```suggestion
type SelectedItemStoreState = DetailsInfo & {
```

see the comment below

src/gui/src/components/explorer-grid/selected-item/context.tsx
```diff
@@ -48,59 +54,103 @@ const getSocketInsights = (
...
+      depIdSpec,
+      abortController.signal,
+      manifest,
+    )) {
+      store.setState(state => ({
+        selectedItemDetails: {
+          ...state.selectedItemDetails,
+          ...d,
+        },
+      }))
```
ruyadorno, line 129: In case you agree with the prior suggestion than we will of course need to change it here, along with the `SelectedItemStoreAction` definition and implementation above, just pointing out it here as a reminder.
```suggestion
      store.setState(state => ({
        ...state.selectedItemDetails,
        ...d,
      }))
```

src/gui/src/components/explorer-grid/selected-item/item-header.tsx
ruyadorno, line 1: as mentioned in https://github.com/vltpkg/vltpkg/pull/682/files#r2055062457 I would expect each portion that relies on specific properties from `fetchDetails` to be isolated as its own component, this way we can be sure there are no extra rerenders for each specific subcomponent.

ruyadorno (changes requested):
I really love the resulting API here that ends up using the zustand store API everywhere as that just normalize a lot of this props/store retrieval.

That said, I have the impression that using both the context API and the zustand store API might be duplication and overhead that is not necessarily needed.

src/gui/src/components/explorer-grid/selected-item/context.tsx
ejkorol: Agreed!

ruyadorno (approved):

PR #683, opened 2025-04-22
gui: bugfix header build version

ejkorol:
Fixes the build version in the `header` component by setting the `buildVersion` in `zustand` and `localStorage`

Closes vltpkg/statusboard#119

- **creates** `buildVersion` type in `zustand` for app version
- **passes** `updateBuildVersion` `zustand` state variable to `start-dashboard-data`
- **updates** `start-dashboard-data` test

### DIFF
--- a/src/gui/src/state/index.ts
+++ b/src/gui/src/state/index.ts
@@ -70,6 +70,9 @@ const initialState: State = {
     )
     return defaultLabels
   })(),
+  buildVersion: localStorage.getItem(
+    'build-version',
+  ) as State['buildVersion'],
 }
 
 /**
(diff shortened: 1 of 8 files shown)


src/gui/src/state/index.ts
ruyadorno, line 1: I don't think it's necessary to save it to `localStorage` - it seems fine to have it just be a simple getter/setter but maybe I'm missing some detail?

ruyadorno (changes requested):

src/gui/src/state/index.ts
ejkorol: If we are on a route which doesn't trigger `startDashboardData()` then the `buildVersion` would never be defined. So if we are on the `/explore` route and refresh the page, then `buildVersion` would be undefined. We would need to set `buildVersion` in a place where it could persist across page refreshes, such as `localStorage`

PR #700, opened 2025-04-24
New and improved exec with install

isaacs:
Exec commands `vlt exec`/`vlt x`/`vlx` will now install using the `@vltpkg/vlx` utility module.

The `vlt exec-cache`/`vlt xc` commands can be used to interact with the installed packages in the XDG data `'vlt/vlx'` directory.

Running `vlx --package=<spec>` without any arguments will place that vlx folder's `node_modules/.bin` first in the `PATH` env, and drop into an interactive shell, so that its bins can be accessed.

As with `npx` and its like, `vlx pkg@version` will run the default command for that package spec, if it can be determined (or raise an error if not).

This also cleans up a lot of the run/exec output so that it's a lot less noisy, and more closely follows the pattern of the other commands, with custom views provided by the `src/exec-command.ts` shared module.

Also, along the way:

- Fixed a bug where `stale-while-revalidate` revalidation was happening in the wrong cache folder, which is not super useful! Now it's fixed.
- Added the `vlt cache info` command, which can be used to print both the cache metadata _and_ the cache body, for inspecting packuments, tarballs, and so on.


src/vlx/src/dir-exists.ts
```diff
@@ -0,0 +1,7 @@
+import { stat } from 'fs/promises'
```
lukekarrys, line 1:
```suggestion
import { stat } from 'node:fs/promises'
```

src/vlx/README.md
```diff
@@ -0,0 +1,76 @@
...
+**[Usage](#usage)**
+
+## Overview
+
+This is a tool that provides the global installs used by `vlt exec`
+(that is, when it's not a local installed dependency).
+
+## Usage
+
+```js
```
lukekarrys, line 14:
```suggestion
```ts
```

Changing this will update some line lengths in the code block once prettier is run.

src/vlx/README.md
```diff
@@ -0,0 +1,76 @@
...
+import * as vlx from '@vltpkg/vlx'
+
+// get the options and arguments in the normal ways
+const conf: LoadedConfig = blahblahblah()
+
+const { positionals, options} = conf
+
+// Provide a promptFn which is ()=>Promise<string> if you want
```
lukekarrys, line 22:
```suggestion
// Provide a promptFn which is `()=>Promise<string>` if you want
```

This is what's causing the docs CI job to fail. Since this all gets parsed as MDX eventually we have to always do MDX type things even in markdown.

src/vlx/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export type VlxManifest = Manifest & {
+  vlx?: {
+    integrity?: Integrity
+    signatures?: Resolution['signatures']
+  }
+}
+
+/**
+ * Provide a promptFn which is ()=>Promise<string> if you want to confirm
```
lukekarrys, line 18:
```suggestion
 * Provide a promptFn which is `()=>Promise<string>` if you want to confirm
```

Same here.

lukekarrys (approved):

PR #701, opened 2025-04-24
cli-sdk: update list & query cmd help

ruyadorno:
Fixes up references to no longer supported class selectors and added an intro to the query language making a reference to the Dependency Selector Syntax name.


lukekarrys (approved):
Code looks good, but needs snapshots updated. Also these might need updating in `www/docs` as well.

lukekarrys (approved):

sins921 (approved):

PR #708, opened 2025-04-25
gui: fix non public registry api calls

ruyadorno:
If a package is not identified as being a registry package that is coming from the public registry, then do not try to make calls to the public registry to retrieve information about it.


ejkorol (approved):
👍

PR #711, opened 2025-04-28
gui: selected item contributors

ejkorol:
Screenshot 2025-04-28 at 14 42 33

renders the `contributors` from the `manifest` in the `tabs-overview`

### DIFF
--- a/src/gui/src/components/explorer-grid/selected-item/tabs-overview.tsx
+++ b/src/gui/src/components/explorer-grid/selected-item/tabs-overview.tsx
@@ -1,4 +1,3 @@
-import { useState, useEffect } from 'react'
 import { TabsTrigger, TabsContent } from '@/components/ui/tabs.jsx'
 import { useSelectedItemStore } from '@/components/explorer-grid/selected-item/context.jsx'
 import Markdown from 'react-markdown'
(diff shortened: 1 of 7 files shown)


src/gui/src/components/explorer-grid/selected-item/tabs-overview.tsx
```diff
@@ -23,6 +36,65 @@ export const OverviewTabButton = () => {
...
+          setAvatarImg(avatar)
+        }
+      }
+    }
+
+    void fetchAvatar()
+  }, [contributor])
+
+  return avatarImg
+}
```
ruyadorno, line 60: The idea of having the `src/gui/src/lib/external-info.ts` helper is precisely to take all this extra logic away from the component and avoid having overly complex components with hooks everywhere.

I think the way to go here is to move this logic to `external-info.ts`, expand the returned `DetailsInfo` object to contain a `contributors` property that returns an array of type `Person & { avatar: ImageInfo}` that gets the contributor info along with the avatar info required.

src/gui/src/components/explorer-grid/selected-item/tabs-overview.tsx
```diff
@@ -133,8 +205,24 @@ export const OverviewTabContent = () => {
         </div>
       }
 
+      {manifest?.contributors && manifest.contributors.length > 0 && (
```
ruyadorno, line 208: Like mentioned above, let's get a `contributors` property in the context store instead of retrieving it directly from the manifest.

ruyadorno (changes requested):

src/gui/src/lib/external-info.ts
```diff
@@ -393,6 +433,9 @@ export async function* fetchDetails(
 
     // retrieve download range for the last year from the registry
     trackPromise(fetchDownloadsLastYear())
+
+    // retrieve contributors from the manifest
+    trackPromise(fetchContributors())
```
ruyadorno, line 438: I think this needs to be limited to public-registry-only packages, it's better to move it to outside this if block

ruyadorno (approved):

PR #713, opened 2025-04-29
gui: selected-item  stars, issue and pr count

ejkorol:
## More Metadata

Screenshot 2025-04-29 at 12 31 56

### New additions:
- creates new style for an `aside` like view for the details of the package
- updates some styles for consistency
- shows the stars, open issue count, open pr count and repository information on the package

### DIFF
--- a/src/gui/src/lib/external-info.ts
+++ b/src/gui/src/lib/external-info.ts
@@ -1,7 +1,11 @@
 import type { Repository, Manifest, Packument } from '@vltpkg/types'
 import { compare, gt } from '@vltpkg/semver'
-import { Spec } from '@vltpkg/spec/browser'
 import { isRecord } from '@/utils/typeguards.js'
+import type { Spec } from '@vltpkg/spec/browser'
+import {
+  getRepoOrigin,
+  getRepositoryApiUrl,
+} from '@/utils/get-repo-url.js'
 
 export type Semver = `${number}.${number}.${number}`
 
@@ -15,6 +19,22 @@ export const asSemver = (s: string): Semver => {
     return s
   }
   throw new Error(`Invalid Semver: ${s}`)
+}
+
+export type GitHubRepo = {
+  owner?: {
+    avatar_url?: string
+    login?: string
+  }
+  updated_at?: string
+  stargazers_count?: number
+  organization?: {
+    login?: string
+  }
+  name?: string
+  default_branch?: string
+  commits_url?: string
+  contributors_url?: string
 }
 
 export type DownloadsRange = {
@@ -66,6 +86,9 @@ export type DetailsInfo = {
   versions?: Version[]
   greaterVersions?: Version[]
   contributors?: Contributor[]
+  stargazersCount?: GitHubRepo['stargazers_count']
+  openIssueCount?: string
+  openPullRequestCount?: string
 }
 
 export const NAME_PATTERN = /^([^(<]+)/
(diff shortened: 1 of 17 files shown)


src/gui/src/lib/external-info.ts
```diff
@@ -158,6 +178,49 @@ export async function* fetchDetails(
...
+    fetch(
+      `https://api.github.com/search/issues?q=repo:${repo.organization?.login}/${repo.name}+type:pr+state:open`,
+      { signal },
+    )
+      .then(res => res.json())
+      .then((res: { total_count?: number }) => ({
+        openPullRequestCount: res.total_count,
+      }))
+      .catch(() => ({}))
+
```
ruyadorno, line 223: We should avoid these extra REST API requests.

For now, I believe the best way to go is to use only the Issues+PRs count (`open_issues_count`) that is already available in the current `fetch(githubAPI)` call and have an adapted UI to present that info. Otherwise we'll be hitting much (3x) faster the 60 unauthenticated requests/hour limit, ref: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users

Long term, if we really believe this is an essential info to display then we should look into using the GraphQL API instead in order to avoid the problem of receiving a rather large payload only to count items. Also, if we believe this is essential info to display then we need to look into ways to authenticate end users so that they're not subject to the very limited unauthenticated rate.

ruyadorno (changes requested):

src/gui/src/lib/external-info.ts
```diff
@@ -202,45 +265,33 @@ export async function* fetchDetails(
...
-                    `${repo.owner.login}'s avatar`
-                  : 'avatar',
-              },
-            }
-          } else {
-            return {}
-          }
+  const fetchFavIcon = (githubAPI: string): Promise<DetailsInfo> => {
+    if (seenFavIconRequests.has(githubAPI)) return Promise.resolve({})
+    seenFavIconRequests.add(githubAPI)
```
ruyadorno, line 270: lol nice catch copilot 😅 

looking at this though it does not make much sense, I think we need to then change the `Set<string>` to a `Map<string, Promise<DetailsInfo>>` in order to return the proper results if needed.

honestly, maybe it's worth revisiting as this `Set` and this check might no longer be needed as there could be other checks happening.

src/gui/src/lib/external-info.ts
ejkorol: I agree that long term the best way would be consolidating this data and fetching from a single endpoint

The way I see it is we have 2 options:
- Utilizing the GraphQL API and allowing end users to authenticate with the GUI
- Or serving this data from our own VSR instance

Pertaining to our offline conversation, I did explore another approach and settled for the hackier way temporarily via  [GitHub shields](https://github.com/badges/shields) until we decide on a plan and strategy for the long term

src/gui/src/lib/external-info.ts
```diff
@@ -202,45 +288,36 @@ export async function* fetchDetails(
...
-            alt: 'avatar',
+            src: repo.owner.avatar_url,
+            alt:
+              repo.owner.login ?
+                `${repo.owner.login}'s avatar`
+              : 'avatar',
           },
         }
-      })
+      }
```
ruyadorno: Can we please rollback this refactor 🙏

src/gui/src/lib/external-info.ts
```diff
@@ -395,6 +473,14 @@ export async function* fetchDetails(
     trackPromise(fetchDownloadsLastYear())
   }
 
+  if (githubRepoPromise) {
+    void githubRepoPromise.then(repo => {
+      trackPromise(fetchStargazerCount(repo))
+      trackPromise(fetchOpenIssuesCount(repo))
+      trackPromise(fetchOpenPullRequestCount(repo))
+    })
+  }
```
ruyadorno, line 483: we should not rely on having the github repo promise resolve (as that can be rate limited), instead what we want to do is extract the orgname and reponame from the `package.json` "repository" property as we have done in the `retrieveGitHubAPIUrl` function above

src/gui/src/lib/external-info.ts
```diff
@@ -137,6 +156,19 @@ export const retrieveAvatar = async (
   return `https://gravatar.com/avatar/${hash}?d=404`
 }
```
ruyadorno, line 157: Let's refactor `retrieveGitHubAPIUrl` to share the same logic to sanitize the github url from src/gui/src/utils/get-repo-url.ts

ruyadorno (approved):

PR #719, opened 2025-04-30
deps: update knip + config

webpro:
Sorry it was hurting my eyes, just had to do this 🤷‍♂️ 😃


lukekarrys (approved):

webpro:
Thanks! And no, it was all on ~me~ Knip!

- `astro.config.mts` and pnpm massacre were fixed this morning (hence the bump)
- the Starlight auto-entries are only available since v5.48.0

Was actually glad to see a nice configuration, but couldn't stand you had to do all this.

PR #724, opened 2025-04-30
Fix manifest name confusion

ruyadorno:
Better handle manifest name confusion in many components of the client:
- `@vltpkg/dep-id` now detects and fixes up an id in its `getTuple` method
- `@vltpkg/graph`
    - properly detect and signals a manifest-confusion when building a new graph
    - added `node.confused` and `node.rawManifests` property to handle that info
    - added to lockfile transfer metadata in order to make info available to the GUI
- `@vltpkg/query` added support to `node.confused` to the `:confused` Insight Selector
- `@vltpkg/cli-sdk` `vlt ls` and `vlt query` will now properly display `confused` metadata in both human and json output
- `@vltpkg/gui` correctly renders the original manifest while the rest of the UI uses fixed up manifest metadata

#### Screenshots

Screenshot 2025-04-30 at 5 21 35 PM
> GUI now properly shows the actual installed package and its manifest data

Screenshot 2025-04-30 at 4 58 51 PM
> `vlt ls` render of the same confused package

Fixes: https://github.com/vltpkg/statusboard/issues/139


lukekarrys (approved):
LGTM!

One question I had: is the plan is to add more manifest fields as markers for the `confused` property? I see this PR uses `isPackageNameConfused` so I was curious if in the future we might add `isPackageVersionConfused` and wrap those both in a `isPackageConfused` check which would set the internal `#confused` flag on a node.

And a minor suggestion: I think the smaller commits that are only updating snapshots in dependent workspaces should be part of the commit making the src changes as well. I find this pattern helpful when doing things like `git bisect` because it means a single commit is more likely to pass all the tests in the repo. I don't always follow this for my PRs so I think of it as a guideline. But after debugging recently and checking out specific commits, I think its a helpful one.

PR #750, opened 2025-05-06
gui: refactor dependency item sidebar

ejkorol:
The first step in creating more complex ui interactions in the `dependency-sidebar` is to refactor the existing code
and breaking it down into smaller components to make the state and actions easier to manage.

Utilizes a similar design pattern to creating a scoped `zustand` store within context to initialize it via props in the react lifecycle and having a cleaner component tree

- creates utility for deep equality checks
- creates context and store for `dependency-sidebar`
- refactors `dependency-sidebar` component with the store
- removes old `dependency-side-bar` component
- refactors `add-dependency` popover component


src/gui/src/components/explorer-grid/dependency-sidebar/actions.ts
ruyadorno, line 1: my idea when I refactored and moved the `src/gui/src/lib/start-dashboard-data.ts` module was to have all of the actions that perform requests living in a common place, so if we're to refactor this now I'd prefer to move it to `src/gui/src/lib/add-remove-dependency.ts` instead.

ruyadorno: bonus: it would also allow components not nested within `components/explorer-grid/dependency-sidebar` to reuse it

ruyadorno: staring a bit more at the code, `changePackage` is an awful function name, `addRemoveDependency` might be a better name but def open to suggestions

ruyadorno (changes requested):
just one small change, otherwise LGTM

ruyadorno (approved):

PR #751, opened 2025-05-06
Lint for import/extensions in docs, gui

lukekarrys:
Also refactors some code to work around instead of ignoring lint rules.


src/tar/src/unpack.ts
```diff
@@ -139,9 +139,7 @@ const unpackUnzipped = async (
     let gex: HeaderData | undefined = undefined
     while (
       offset < buffer.length &&
-      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-      (h = new Header(buffer, offset, ex, gex)) &&
-      !h.nullBlock
+      ((h = new Header(buffer, offset, ex, gex)), !h.nullBlock)
```
isaacs, line 142:
```suggestion
      (!(h = new Header(buffer, offset, ex, gex)).nullBlock)
```

isaacs (approved):
Rubberstamp LGTM, seems like obvious improvement.

PR #753, opened 2025-05-06
cli-sdk: reuse jackspeak code for printing config errors

lukekarrys:


src/cli-sdk/src/index.ts
```diff
@@ -50,55 +51,36 @@ const loadCommand = async <T>(
...
+        `Problem in Config File ${path}`
+      : 'Invalid Option Flag'
+    const validOptions =
+      wanted ? undefined
+      : isConfigFile ? getSortedKeys()
+      : getSortedCliOptions()
+    stderr(msg)
+    stderr(err.message)
+    if (name) stderr(indent(`Field: ${format(name)}`))
+    if (found) stderr(indent(`Found: ${format(found)}`))
```
isaacs, line 72:
```suggestion
    if (found) {
      stderr(indent(`Found: ${isConfigFile ? JSON.stringify(found) : format(found)}`))
    }
```

src/cli-sdk/test/print-err.ts
```diff
@@ -189,8 +183,8 @@ t.test('error with an unknown code', t => {
   })
   t.matchStrict(printed, [
     'Error: this is an error',
+    'Code: ENOTACODEWEKNOWABOUT',
     'Cause:',
-    `  code: ENOTACODEWEKNOWABOUT`,
```
isaacs, line 193: Why did this move up to the top level?

src/cli-sdk/src/print-err.ts
```diff
@@ -51,52 +73,40 @@ export const printErr = (
...
-  // We don't know what this is, just print it.
-  stderr(`Unknown Error:`, format(err))
+  if (fileWritten) {
+    stderr('')
+    stderr(`Full details written to: ${fileWritten}`)
+  }
+  if (!knownError || knownError.bug) {
+    stderr('')
+    stderr('Open an issue with the full error details at:')
+    stderr(indent('https://github.com/vltpkg/vltpkg/issues/new'))
```
isaacs, line 97: It'd be good to make this an ANSI hyperlink to the URL that has a title, label, and details already injected into the body, if we can.

https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/creating-an-issue#creating-an-issue-from-a-url-query

isaacs: If the contents are too big to put on the query string, we could just have a section that says "Paste your error log file here" or something.

isaacs (approved):
I think the types probably were better in error-cause. Though it doesn't make much of a difference, and I wouldn't block this PR for it, it feels like `@vltpkg/types` is getting a bit commingled with functionality and definitions that belong to other modules. Ie, for `Manifest` or something that's truly abstract and not really "owned" by any particular piece, it makes sense, but I wonder if the Error and cause object stuff really meets that bar, or if it should be considered to be owned by `@vltpkg/error-cause`. Anyway, more a comment and thing to keep tracking in case it starts to trend in a messy code smell kind of direction, rather than a specific objection. LGTM, I like the outputs and the behavior of dropping a file with all the deets.

src/cli-sdk/src/print-err.ts
isaacs: Oh! another possible trick, if we're on macOS, we can pipe the details into the `pbcopy` program, and then tell them to paste into an issue.

PR #756, opened 2025-05-07
cli: improve how invalid options are reported

isaacs:
This applies the same `code: 'ECONFIG'` behavior for all config options, including invalid values, providing more assistance beyond just listing the available options.

Now it properly detects for example when an option is given a value that shouldn't be (because it's a flag) or vice versa (not given a value to an option that is not a boolean flag), or given an invalid value (eg, `--config=asdfasdf` where only specific values are allowed).

It also properly differentiates between config _file_ errors and cli option errors, presenting the suggestions in a more useful manner.

### DIFF
--- a/src/cli-sdk/src/commands/config.ts
+++ b/src/cli-sdk/src/commands/config.ts
@@ -1,18 +1,19 @@
-import { error } from '@vltpkg/error-cause'
+import type { ErrorWithCauseObject } from '@vltpkg/error-cause'
+import { error, isErrorWithCause } from '@vltpkg/error-cause'
 import { spawnSync } from 'node:child_process'
-import {
-  definition,
-  isRecordField,
-  pairsToRecords,
-  recordsToPairs,
-} from '../config/index.ts'
+import { getSortedKeys } from '../config/definition.ts'
 import type {
   ConfigDefinitions,
   ConfigFileData,
   LoadedConfig,
   RecordPairs,
 } from '../config/index.ts'
-import { getSortedKeys } from '../config/definition.ts'
+import {
+  definition,
+  isRecordField,
+  pairsToRecords,
+  recordsToPairs,
+} from '../config/index.ts'
 import { commandUsage } from '../config/usage.ts'
 import type { CommandFn, CommandUsage } from '../index.ts'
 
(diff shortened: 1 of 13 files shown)


ruyadorno (approved):

src/cli-sdk/test/config/definition.ts
```diff
@@ -1,4 +1,6 @@
 import t from 'tap'
+import { getSortedCliOptions } from '../../src/config/definition.ts'
```
lukekarrys, line 2: This is why coverage is being detected incorrectly. All imports for a file need to be done with `t.mockImport` if there is top-level state being altered in the tests: https://github.com/vltpkg/vltpkg/blob/main/CONTRIBUTING.md#test-coverage-is-failing-but-it-shouldnt-be

src/cli-sdk/src/commands/config.ts
```diff
@@ -191,12 +192,47 @@ const set = async (conf: LoadedConfig) => {
...
   try {
     parsed = conf.jack.parseRaw(pairs.map(kv => `--${kv}`)).values
-  } catch {
-    throw error('Invalid config keys', {
-      code: 'ECONFIG',
-      found: pairs.map(kv => kv.split('=')[0]),
-      wanted: getSortedKeys(),
-    })
+  } catch (er) {
+    const { cause } = er as ErrorWithCauseObject
```
lukekarrys, line 196: I think it's better to use the `isErrorWithCause` type guard from `error-cause` here and have it just `throw er` at the end if its not. The only reason is to protect against head-scratching dev time errors where the `try` block throws for some other reason.

src/cli-sdk/src/index.ts
```diff
@@ -47,25 +50,49 @@ const loadCommand = async <T>(
...
+      const validOptions =
+        wanted ? undefined
+        : isConfigFile ? getSortedKeys()
+        : getSortedCliOptions()
+
+      stderr(msg)
+      stderr(err.message)
+      if (name) stderr(indent('Field: ' + name))
+      if (found) stderr(indent('Found: ' + JSON.stringify(found)))
+      if (wanted) stderr(indent('Wanted: ' + wanted))
```
lukekarrys, line 85: If we are in this block it means we have no config so we can't use the same formatting options we use with `util.formatWithOptions`. But we should still use some sort of format function here even if its just `util.format`. This will fix the linting errors.

lukekarrys (approved):
These error messages are much better. Left a few comments but nothing blocking. So once CI is green :shipit:

src/cli-sdk/src/index.ts
isaacs: At this point we know it's coming from Jackspeak, and thus `wanted` is always a string, so I fixed it by just checking `if (typeof wanted === 'string')` which future proofs it and also satisfies lint.

src/cli-sdk/src/commands/config.ts
isaacs: Right, but then we still don't know that `cause` is a cause _object_, just that it's not undefined.

I can still add the guard, but the `as ErrorWithCauseObject` will have to stay, unless we add a `isErrorWithCauseObject` to `@vltpkg/error` (which isn't a terrible idea, of course).

PR #757, opened 2025-05-07
spec: default npm: alias to configured default registry

isaacs:
Based on #756 land that first

If you're setting your default registry to a caching proxy, it'll be
very surprising to have any public-registry packages using the npm:
alias to be fetched from the public registry, which your system might
not even be able to access.

This fixes the problem by making the npm: alias default to the
*configured* default registry (which of course defaults to the default
public registry), so there's no change in the normal case, but it'll
stop being broken in the cause where the public registry is pointed
somewhere else, without requiring the user to *also* set the npm: alias
when changing the default registry.

Found while writing documentation about working with registries.


src/gui/test/__snapshots__/routes.tsx.snap
```diff
@@ -85,8 +85,6 @@ exports[`renders Layout for the "/help" view 1`] = `
     <gui-sidebar-inset>
       <gui-header>
       </gui-header>
-      <gui-help-selectors>
-      </gui-help-selectors>
```
ruyadorno, line 89: is this change intentional? 🤔

isaacs: That happened when I ran tests at the top level, I'm trying to figure out why it happened and why it's failing.

But no.

src/spec/src/browser.ts
```diff
@@ -408,6 +400,8 @@ export class Spec implements SpecLike<Spec> {
       }
     }
 
+    const regs = Object.entries(this.options.registries)
+    regs.push(['npm', this.options.registry])
```
ruyadorno, line 404: what if `this.options.registries.npm` is already defined? should there be a condition here for that?

isaacs: No. If it's already defined, it'll come before this one, and this will be a no-op. It's only relevant if no other setting has been defined.

isaacs: But may as well have a guard against the unnecessary push, I guess. Updated.

lukekarrys (approved):

PR #761, opened 2025-05-07
graph: add node first edge in getter

ruyadorno:
Adds a new `node.firstEdgeIn` getter that retrieves a single item that represents a link to a deterministic "parent" node, which will make it possible to have transitive nodes and edges trace a unique path to the graph importer.

Refs: https://github.com/vltpkg/statusboard/issues/133


ruyadorno:
This approach is not going to work.

PR #768, opened 2025-05-12
bugfix: docs theme switcher

ejkorol:
Fixes vltpkg/statusboard#131

- fix theme provider to properly handle missing theme cookie by defaulting to 'system+dark'
- fix theme switcher component to default to 'system' when no theme cookie exists
- ensure consistent theme handling to prevent undefined states


lukekarrys (approved):
Code looks good and I ran it locally. It's hard to test in PR deployments because I think we only allow cookies on `.vlt.sh` domains when not in development. But that's not a blocker to this.

PR #769, opened 2025-05-13
gui bugfix: header build version

ejkorol:
### Bug Explanation

When the gui initially loads, it makes a request via `startDashboardData` on routes that use that endpoint. This function sets the `buildVersion` in the zustand store.
However there is a bug where if you do not start on a route that calls `startDashboardData`, then `buildVersion` is undefined. I.e. opening the gui in a project.

Or when performing a hard refresh on the `explore` route, the `buildVersion` is undefined because the `startDashboardData` endpoint is not called and the zustand `buildVersion` is not set.

### Bug Fix

We create a common new endpoint `./app-data.json` that serves data which the gui needs on all endpoints.

We refactor the `start-dashboard-data` lib file to `start-data` to be more generic, and we create a generic `requestData` function that will be used to fetch data from endpoints.

This endpoint is called in the `usePreflight` hook inside of `layout.tsx`, which is used in the zustand store to set the initial state of the app.

This ensures that every route that uses the `usePreflight` hook will have the `buildVersion` set in the zustand store.

Closes vltpkg/statusboard#119

#### Changes Made
- **create new `./app-data.json` endpoint in `server`**
- **updates tests for `app-data.ts` and `dashboard.ts`**
- **adds `AppData` into zustand state**
- **updates `start-dashboard-data` to `start-data`**
- **creates `use-preflight` hook**
- **updates `create-new-project` starting data**
- **updates `dashboard` start data**
- **updates `queries` start data**
- **mocks the `usePreflight` hook for tests**
- **updates dashboard mocks in `directory-select`**
- **updates `header` component for new `appData`**

### DIFF
--- a/src/server/test/app-data.ts
+++ b/src/server/test/app-data.ts
@@ -19,20 +19,6 @@ t.test('AppDataManager construction', async t => {
   t.equal(manager.publicDir, publicDir)
 })
 
-t.test('update app data', async t => {
-  const dir = t.testdir({
-    public: {},
-  })
-  const publicDir = resolve(dir, 'public')
-  const manager = new AppDataManager({ publicDir })
-
-  t.equal(await manager.update(), true)
-  const appDataPath = resolve(dir, 'public/app-data.json')
-  t.matchOnly(JSON.parse(readFileSync(appDataPath, 'utf8')), {
-    buildVersion: '1.2.3',
-  })
-})
-
 t.test('update app data with existing file', async t => {
   const dir = t.testdir({
     public: {


src/server/test/app-data.ts
```diff
@@ -0,0 +1,52 @@
...
+  })
+  const publicDir = resolve(dir, 'public')
+  const manager = new AppDataManager({ publicDir })
+
+  t.equal(await manager.update(), true)
+  const appDataPath = resolve(dir, 'public/app-data.json')
+  t.matchOnly(JSON.parse(readFileSync(appDataPath, 'utf8')), {
+    buildVersion: '1.2.3',
+  })
+})
```
ruyadorno, line 34: if this gives no extra coverage I think it's better to remove it

ruyadorno (approved):

PR #773, opened 2025-05-14
fix: Handle jsr: without a name like yarn, pnpm do

isaacs:
That is, `@a/b@jsr:1.x` should be the same as `@a/b@jsr:@a/b@1.x`.

### DIFF
--- a/src/graph/tap-snapshots/test/lockfile/save.ts.test.cjs
+++ b/src/graph/tap-snapshots/test/lockfile/save.ts.test.cjs
@@ -90,7 +90,7 @@ exports[`test/lockfile/save.ts > TAP > jsr-registries > must match snapshot 2`] 
     ]
   },
   "edges": {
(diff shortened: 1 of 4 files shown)


ruyadorno:
Why the alias-less protocol should work only for `jsr:` and not for any custom protocol?

ruyadorno:
I see! 

I guess that _maybe_ we could tweak the current implementation so that if a `subspec` is missing a name it could inherit it, e.g: `foo@npm:@1` since we'd still need the `@` in order to disambiguate.

Not sure if it's an improvement and def not something we need to address on this PR but I wanted to hear your thoughts on it.

isaacs:
That at least conceivably could work, because scoped packages must have a `/`, and `/` isn't valid in dist-tags or version ranges.

But it's also a departure from the `npm:` alias style, so I wonder if it's worth inviting the incompatibility, if we'd want to prevent `npm:` aliases specifically from being unnamed. Doesn't seem like it gets us much more than just added complexity and a weird edge case.

ruyadorno (approved):

PR #775, opened 2025-05-14
gui: separate grid results and selected item components

lukekarrys:
There should be no behavior or visual changes in this PR. The `<ExplorerGrid>` is now a smaller wrapper which will get the list of items from the query and either render a `<SelectedItem>` or a `<Results>`. Both of these new components require the data to be passed now via props instead of getting it from the store.

This is the basis to then make changes that will only be in `<Results>`. Ref https://github.com/vltpkg/statusboard/issues/111


src/gui/src/components/explorer-grid/selected-item.tsx
ruyadorno, line 1: the naming here is really confusing since `src/gui/src/components/explorer-grid/selected-item/index.ts` used to have this name.

One idea, from @ejkorol is to nest both this file and `results.tsx` inside a `results/` folder that then can also have `empty-results-state.tsx` in it (and then maybe rename `results.tsx` to `index.tsx` and `selected-item.tsx` to `item.tsx`). wdyt?

lukekarrys: Good call. I made all the changes and we know have nice and clean `results/` and `selected-item/` directories with their relevant component files.

Issue #638, opened 2025-04-10
[BUG] `install` will remove explicitly defined dist-tags in `package.json`


darcyclarke:
### Expected Behaivour
- `vlt install` respects `package.json` dependency definitions & leaves them alone

### Current Behaivour
- `vlt install` will rewrite `package.json` dependency definitions so that any dist-tags are converted to caret semver ranges

### Reproduction Steps

1. `vlt init`
1. `vlt pkg set dependencies.express=latest`
1. `vlt pkg get dependencies.express # latest`
1. `vlt install`
1. `vlt pkg get dependencies.express # ^5.1.0`

PR #778, opened 2025-05-15
reify: save properly back to package.json

isaacs:
This fixes a few overlooked issues in how we save dependencies back to package.json. The new logic is, in a nutshell:

- If the user provided a specific version or range in the installation, we save that back to package.json
- If the package.json specified a specific single version, then we save that (ie, keep it pinned).
- If the final registry alias or jsr specifier did not include a version, and the package.json _does_ include a version or range, leave it un-touched, because we were just installing what the package.json said to, and it remains the authority.
- If the final registry alias or jsr specifier did not include a version range, and the package.json does not include a version or range, then we use the `^{resulting version}` range.

Fix: #638
Fix: VLT-344

### DIFF
--- a/src/graph/test/reify/calculate-save-value.ts
+++ b/src/graph/test/reify/calculate-save-value.ts
@@ -1,0 +1,80 @@
+import { Spec } from '@vltpkg/spec'
+import t from 'tap'
+import { calculateSaveValue } from '../../src/reify/calculate-save-value.ts'
+
+t.test('non-registry type, just saved as-is', t => {
+  const s = Spec.parse('x', 'github:a/b')
+  t.equal(calculateSaveValue('git', s, 'a/b', '1.2.3'), 'github:a/b')
+  t.end()
+})
+
+// registry cases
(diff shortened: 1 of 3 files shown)


src/graph/test/reify/calculate-save-value.ts
```diff
@@ -0,0 +1,73 @@
...
+// registry cases
+// nodeVersion is always 1.2.3
+const cases: [
+  spec: string,
+  existing: string | undefined,
+  expect: string,
+  comment: string,
+][] = [
+  ['1', undefined, '1', 'new dep, use specified range'],
+  ['', undefined, '^1.2.3', 'new dep, no spec, use caret range'],
```
ruyadorno, line 20: maybe add a few more variety of ranges, even though it should be the same code path I think it might guard against future regressions, I mean things like `>2.1 || 3`, `~1.1.0`, `1.x.x`

ruyadorno (approved):
LGTM

PR #787, opened 2025-05-22
gui: dependency item design + filtering

ejkorol:
- **installs `radix-ui/react-context-menu`**
- **updates `badge` styles**
- **updates `data-badge` styles**
- **updates `dropdown-menu` styles**
- **adds `xxs` font size tailwind**
- **updates `dependency-sidebar` context for filters**
- **updates `dependency-sidebar/index.tsx`**
- **updates `add-dependency` button styles and test**
- **creates `filters` for `dependency-sidebar`**
- **updates `explorer-grid` header styles**
- **updates `label-helper` label styles**
- **updates `side-item` styles**
- **removes `lineRef` state (no longer needed)**
- **updates the `selected-item` view**
- **removes the old connector line logic**
- **updates `grid-header` styles on result**

### DIFF
--- a/src/gui/src/components/explorer-grid/dependency-sidebar/context.tsx
+++ b/src/gui/src/components/explorer-grid/dependency-sidebar/context.tsx
@@ -11,10 +11,14 @@ import type { PropsWithChildren } from 'react'
 import type { GridItemData } from '@/components/explorer-grid/types.ts'
 import type { DependencySideBarProps } from '@/components/explorer-grid/dependency-sidebar/index.tsx'
 import type { Operation } from '@/lib/add-remove-dependency.ts'
+import type { Filter } from './filter-config.ts'
 
 type DependencySidebarState = {
   importerId?: DependencySideBarProps['importerId']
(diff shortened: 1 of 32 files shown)


src/gui/src/components/explorer-grid/dependency-sidebar/context.tsx
```diff
@@ -106,6 +128,20 @@ export const DependencySidebarProvider = ({
...
+    }))
+  }, [dependencies])
+
+  /** Reset the filters when navigating */
+  useEffect(() => {
+    dependencySidebarStore.setState(() => ({
+      filters: [],
+    }))
+  }, [query])
+
```
lukekarrys, line 144: This seems like it goes against the rule of react that you should not update state based on other state/props inside an effect: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

I'm not exactly sure how it would be refactored without looking closer at how `dependencySidebarStore` is used through the component tree. But in general it seems like we should be able to either calculate the filtered set of dependencies during render or if that is expensive do it inside `useMemo` instead.

PR #793, opened 2025-05-27
chore: update to pnpm@10

lukekarrys:
This is mainly so we can use the jsr: prefix and remove the .npmrc file


www/docs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@vltpkg/docs",
-  "version": "0.0.0",
+  "version": "0.0.0-1",
```
lukekarrys, line 3: This is to trigger a docs PR deploy to make sure `pnpm@10` works in the Vercel build.

isaacs (approved):
LGTM

cceneag (approved):

PR #794, opened 2025-05-28
add @vltpkg/vlt-json

isaacs:
This change is fairly consequential. It makes for some convenient affordances at run-time, but a few somewhat _inconvenient_ caveats at dev-time.

Basically, _any_ time a section is read from the `vlt.json` file, it is cached. The location of the `vlt.json` file is cached. Everything is memoized to make it virtually _impossible_ to accidentally have to read the file multiple times, overwrite another's use of it, or search for it more than once.

The downside is, of course, that's _exactly_ what we want to repeatedly in tests.

To account for this, a `reload(<section>, <whichconfig>)` method is added, as well as a `unload(<whichConfig>)` method. These should _not_ be used at run time, but are affordances for tests to be able to reload the configuration rather than used the memoized copy of the data object. Note that _many_ tests had to be updated to add a `vlt.json` file into the `t.testdir()` call, and then `t.chdir(projectRoot); unload('project')` to get it to forget about its previous configuration (and not fall up to the actual `vlt.json` file in the root of this repo!)

All in all, I feel like it's a worthwhile trade-off, even though it does require some care when writing tests that load configuration. Perhaps we could have it _not_ memoize configuration when it detects that it's running in a test or something, but that seems like it's probably a bit too clever.


src/cli-sdk/package.json
```diff
@@ -39,19 +39,20 @@
...
     "jackspeak": "^4.1.0",
     "package-json-from-dist": "catalog:",
     "path-scurry": "catalog:",
     "polite-json": "catalog:",
     "pretty-bytes": "^6.1.1",
-    "react": "^18.3.1",
     "react-devtools-core": "^4.28.5",
+    "react": "^18.3.1",
     "supports-color": "^10.0.0",
     "walk-up-path": "catalog:"
```
lukekarrys, line 57: `walk-up-path` is no longer being used in `cli-sdk` so i think this can be removed

lukekarrys (approved):
LGTM.

Two minor changes to unused deps in package.json. Can be part of this PR or a follow-up.

ruyadorno:
> To account for this, a `reload(<section>, <whichconfig>)` method is added, as well as a `unload(<whichConfig>)` method. These should _not_ be used at run time, but are affordances for tests to be able to reload the configuration rather than used the memoized copy of the data object.

Please note that these are very likely to be used at runtime once we have an editable Config GUI view.

Issue #797, opened 2025-05-28
[BUG] EEXIST: file already exists, symlink during install


lukekarrys:
Found this as part of #792. To reproduce follow instructions at https://github.com/vltpkg/projects-that-error/blob/main/projects/vltpkg-797/README.md

---

**Actual Output**

```
$ vlt i
build ✓ > actual ✓ > reify⠋
202 requests
Error: EEXIST: file already exists, symlink '../sass-embedded/dist/bin/sass.js' -> '/Users/lukekarrys/projects/vltpkg/projects-that-error/projects/vltpkg-797/node_modules/.vlt/··vite@6.3.5/node_modules/.bin/sass'

Full details written to: /Users/lukekarrys/.local/data/vlt/error-logs/error-79148.log

Open an issue with the full error details at:
  https://github.com/vltpkg/vltpkg/issues/new
```

Error log
```
Error: EEXIST: file already exists, symlink '../sass-embedded/dist/bin/sass.js' -> '/Users/lukekarrys/projects/vltpkg/projects-that-error/projects/vltpkg-797/node_modules/.vlt/··vite@6.3.5/node_modules/.bin/sass'
    at async symlink (node:internal/fs/promises:1005:10)
    at async Promise.all (index 1)
    at async addEdge (/Users/lukekarrys/.config/packages/src/graph/src/reify/add-edge.ts:75:24)
    at async Promise.all (index 6)
    at async reify_ (/Users/lukekarrys/.config/packages/src/graph/src/reify/index.ts:116:27)
    at async reify (/Users/lukekarrys/.config/packages/src/graph/src/reify/index.ts:61:5)
    at async install (/Users/lukekarrys/.config/packages/src/graph/src/install.ts:29:16)
    at async command (/Users/lukekarrys/.config/packages/src/cli-sdk/src/commands/install.ts:25:21)
    at async outputCommand (/Users/lukekarrys/.config/packages/src/cli-sdk/src/output.ts:149:33)
    at async Module.run (/Users/lukekarrys/.config/packages/src/cli-sdk/src/index.ts:114:3) {
  errno: -17,
  code: 'EEXIST',
  syscall: 'symlink',
  path: '../sass-embedded/dist/bin/sass.js',
  dest: '/Users/lukekarrys/projects/vltpkg/projects-that-error/projects/vltpkg-797/node_modules/.vlt/··vite@6.3.5/node_modules/.bin/sass',
  cause: {
    errno: -17,
[...]

Issue #796, opened 2025-05-28
[BUG] astro check and sync command fail after self install


lukekarrys:
Found this as part of #792. To reproduce see https://github.com/vltpkg/projects-that-error/blob/main/projects/vltpkg-796/README.md

**Actual Output**

```
╰❯ vlt i
build ✓ > actual ✓ > reify⠦
2509 requests
Error: failed graph traversal

Full details written to: /Users/lukekarrys/.local/data/vlt/error-logs/error-4994.log

Open an issue with the full error details at:
  https://github.com/vltpkg/vltpkg/issues/new
```

Error log

```
Error: failed graph traversal
    at async build2 (/Users/lukekarrys/.config/packages/src/graph/src/reify/build.ts:27:3)
    at async reify_ (/Users/lukekarrys/.config/packages/src/graph/src/reify/index.ts:118:3)
    at async reify (/Users/lukekarrys/.config/packages/src/graph/src/reify/index.ts:60:5)
    at async install (/Users/lukekarrys/.config/packages/src/graph/src/install.ts:29:16)
    at async command (/Users/lukekarrys/.config/packages/src/cli-sdk/src/commands/install.ts:25:21)
    at async outputCommand (/Users/lukekarrys/.config/packages/src/cli-sdk/src/output.ts:150:33)
    at async Module.run (/Users/lukekarrys/.config/packages/src/cli-sdk/src/index.ts:114:3)
    at async run (/Users/lukekarrys/.config/packages/npm/build/src/bins.ts:30:5)
    at async <anonymous> (/Users/lukekarrys/.config/packages/npm/build/src/bins/vlt.ts:3:1) {
  [cause]: {
    node: _Node [@vltpkg/graph.Node] {
      confused: false,
      edgesIn: [...],
      edgesOut: [...],
      id: 'workspace·www§docs',
      importer: true,
      mainImporter: false,
      graph: [...],
      integrity: undefined,
      manifest: {
        name: '@vltpkg/docs',
        version: '0.0.0',
        private: true,
        repository: {
          type: 'git',
          url: 'git+https://github.com/vltpkg/vltpkg.git',
          directory: 'www/docs'
        },
        dependencies: {
          '@astrojs/check': '^0.9.4',
          '@astrojs/mdx': '^4.2.5',
          '@astrojs/react': '^4.2.5',
          '@astrojs/starlight': '^0.30.6',
          '@astrojs/starlight-tailwind': '^3.0.1',
[...]

Issue #795, opened 2025-05-28
[BUG] Cannot find module 'is-extglob' during prepare lifecycle script


lukekarrys:
Found this as part of #792. To reproduce:

```
git checkout lk/dogfood
rm -rf node_modules/ {src,infra,www}/*/node_modules/
vlt i
```

**Actual Output**

```
╰❯ vlt i
build ✓ > actual ✓ > reify⠇
2016 requests
Error: failed graph traversal

Full details written to: /Users/lukekarrys/.local/data/vlt/error-logs/error-99813.log

Open an issue with the full error details at:
  https://github.com/vltpkg/vltpkg/issues/new
```

Error log

```
[Error: failed graph traversal] {
  [cause]: {
    node: _Node [@vltpkg/graph.Node] {
      confused: false,
      edgesIn: [...],
      edgesOut: [...],
      id: 'workspace·src§gui',
      importer: true,
      mainImporter: false,
      graph: [...],
      integrity: undefined,
      manifest: {
        name: '@vltpkg/gui',
        description: 'Look under the hood of a vlt install in HD',
        version: '0.0.0-13',
        repository: {
          type: 'git',
          url: 'git+https://github.com/vltpkg/vltpkg.git',
          directory: 'src/gui'
        },
        dependencies: {
          '@radix-ui/react-avatar': '^1.1.7',
          '@radix-ui/react-checkbox': '^1.2.3',
          '@radix-ui/react-collapsible': '^1.1.8',
          '@radix-ui/react-dialog': '^1.1.11',
          '@radix-ui/react-dropdown-menu': '^2.1.12',
          '@radix-ui/react-label': '^2.1.4',
          '@radix-ui/react-popover': '^1.1.11',
          '@radix-ui/react-scroll-area': '^1.2.6',
          '@radix-ui/react-select': '^2.2.2',
          '@radix-ui/react-separator': '^1.1.4',
          '@radix-ui/react-slot': '^1.2.0',
          '@radix-ui/react-tabs': '^1.1.9',
          '@radix-ui/react-toast': '^1.2.11',
          '@radix-ui/react-tooltip': '^1.2.4',
          '@tanstack/react-table': '^8.21.3',
          '@vltpkg/dep-id': 'workspace:*',
          '@vltpkg/graph': 'workspace:*',
          '@vltpkg/query': 'workspace:*',
          '@vltpkg/security-archive': 'workspace:*',
          '@vltpkg/spec': 'workspace:*',
[...]

PR #798, opened 2025-05-28
lk/configurable catalogs

lukekarrys:
- **chore: update to pnpm@10**
- **dev: make catalog usage configurable and turn them off**


lukekarrys:
Nothing is lost here since we can turn it back on by flipping a switch.

I want to make the errors I'm seeing by dogfooding (#795 #796 #797) easier to reproduce on main, so it would be helpful to have this for me as soon as possible.

isaacs:
Catalogs now supported as of #800.

PR #800, opened 2025-05-28
Add `catalog:` support

isaacs:
Just use it like any other spec, anywhere specs are used.

The cli automatically pulls it in from the `vlt.json` file if set there, same format as pnpm's `pnpm-workspaces.yaml`, but json rather than yaml.

`'catalog'` specs have a subspec which is the actual resolved value.

`catalog` and `catalogs` just get plopped onto the same `config.options` object we pass around everywhere, as part of SpecOptions.


isaacs:
> LGTM once green.
> 
> Looks like the GUI is failing due to missing the catalog types.

Yeah, it was also not saving catalogs to the lockfile, that's important, since it's needed when loading the graph back up again.

PR #811, opened 2025-05-29
Isaacs/better build script error reporting

isaacs:
Based on #803, land that first.

Improves the output handling of build script lifecycle errors that occur during the installation. This is much easier to debug than just "Graph traversal failed", since the actual (colored!) output of the failed script is also printed to the terminal, along with information about what failed and the path to it in the graph.


lukekarrys (approved):
LGTM (once rebased to drop the astro pin commit)

Issue #819, opened 2025-06-03
[BUG] `vlt exec <git repo>` fails to install/execute bins


darcyclarke:
### Expected Behaivour
- `vlt exec` successfully installs git/github repos as packages & executes the corresponding bin

### Current Behaivour
- `vlt exec` fails to install/execute a package when referencing a git repo
- an undefined/dep name error is thrown
- this corrupts the cache entry (ie. the cache key can't be used/referenced to remove it) requiring you to blow away the entire cache to remove the entry (ie. `vlt exec-cache delete` vs. `vlt exec-cache <key>`) - not a major headache but an unfortunate sideffect

### Resources

<img width="949" alt="Image" src="https://github.com/user-attachments/assets/a304301e-a35f-44b6-be5f-0d52ef2a6b84" />

<img width="1714" alt="Image" src="https://github.com/user-attachments/assets/db0ae8a0-f28e-4cfe-939c-23e797a78b1c" />

<img width="355" alt="Image" src="https://github.com/user-attachments/assets/5e7dbc58-887b-4427-9861-a66b97bbcd70" />

PR #821, opened 2025-06-03
vlx: infer a reasonable name from nameless spec

isaacs:
Fix: #819


src/vlx/src/install.ts
```diff
@@ -1,7 +1,7 @@
 import { error } from '@vltpkg/error-cause'
 import { install } from '@vltpkg/graph'
 import { PackageInfoClient } from '@vltpkg/package-info'
-import { Spec } from '@vltpkg/spec'
+import type {Spec} from '@vltpkg/spec';
```
lukekarrys:
```suggestion
import type { Spec } from '@vltpkg/spec'
```

PR #822, opened 2025-06-03
graph: add graph modifiers module

ruyadorno:
Add the module that is responsible for loading and applying user-defined "modifiers" loaded from a vlt.json file into a graph at traversal time.

Instances of `GraphModifier` class can be used as a helper to modify the graph during the graph build ideal traversal time. e.g:

```ts
  const modifier = new GraphModifier(options)
  modifier.load(options)

  // The `tryImporter` method can be used to register the initial
  // importer node along with any modifier that includes an importer
  // selector, e.g:
  modifier.tryImporter(graph.mainImporter)

  // When traversing the graph, use the `tryNewDependency` method to
  // check if a given dependency name to the current traversed node
  // has matching registered modifiers, e.g:
  const entries = modifier.tryNewDependency(fromNode, depName)

  // Use `updateActiveEntry` to update a given active modifier entry
  // state with the current node of the graph being traversed. e.g:
  for (const entry of entries)
    modifier.updateActiveEntry(fromNode, entry)
```

### DIFF
--- a/src/graph/src/modifiers.ts
+++ b/src/graph/src/modifiers.ts
@@ -1,6 +1,5 @@
 import { parseBreadcrumb } from '@vltpkg/dss-breadcrumb'
 import { error } from '@vltpkg/error-cause'
-import { PackageJson } from '@vltpkg/package-json'
 import { Spec } from '@vltpkg/spec'
 import { asManifest, assertManifest, isObject } from '@vltpkg/types'
 import { load } from '@vltpkg/vlt-json'
@@ -175,10 +174,6 @@ export type ModifierActiveEntry = {
  */
 export type GraphModifierOptions = {
   /**
-   * A {@link PackageJson} object, for sharing manifest caches
-   */
-  packageJson?: PackageJson
-  /**
    * Options for the {@link Spec} parser
    */
   specOptions?: SpecOptions
@@ -216,8 +211,6 @@ export type GraphModifierOptions = {
  * ```
  */
 export class GraphModifier {
-  /** The package.json object, for sharing manifest caches */
-  packageJson: PackageJson
   /** The loaded modifiers configuration */
   #config?: GraphModifierConfigObject
   /** A set of all modifiers loaded from vlt.json */
@@ -244,14 +237,12 @@ export class GraphModifier {
     ModifierEntry,
     Map<string, Map<Node, ModifierActiveEntry>>
   >()
-
   /**
    * A set of currently active modifiers, which are being parsed.
    */
   activeModifiers = new Set<ModifierActiveEntry>()
 
   constructor(options: GraphModifierOptions) {
-    this.packageJson = options.packageJson ?? new PackageJson()
     this.load(options)
   }
 
--- a/src/graph/src/modifiers.ts
+++ b/src/graph/src/modifiers.ts
@@ -1,0 +1,527 @@
+import { parseBreadcrumb } from '@vltpkg/dss-breadcrumb'
+import { error } from '@vltpkg/error-cause'
+import { PackageJson } from '@vltpkg/package-json'
+import { Spec } from '@vltpkg/spec'
+import { asManifest, assertManifest } from '@vltpkg/types'
+import { load } from '@vltpkg/vlt-json'
+import type {
+  ModifierBreadcrumb,
+  ModifierInteractiveBreadcrumb,
+} from '@vltpkg/dss-breadcrumb'
+import type { SpecOptions } from '@vltpkg/spec'
+import type { Manifest } from '@vltpkg/types'
+import type { Edge } from './edge.ts'
+import type { Node } from './node.ts'
+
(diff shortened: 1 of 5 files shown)


src/graph/test/modifiers.ts
```diff
@@ -0,0 +1,1097 @@
...
+// Graph fixtures code
+const specOptions: SpecOptions = {
+  registries: {
+    custom: 'http://example.com',
+  },
+}
+
+const projectRoot = '.'
+
+export const newGraph = (rootName: string): GraphLike => {
```
lukekarrys:
```suggestion
const newGraph = (rootName: string): GraphLike => {
```

src/graph/test/modifiers.ts
```diff
@@ -0,0 +1,1097 @@
...
+  mainImporter.graph = graph
+  graph.importers = new Set([mainImporter])
+  graph.mainImporter = mainImporter
+  graph.nodes = new Map([[mainImporter.id, mainImporter]])
+  graph.edges = new Set()
+
+  return graph
+}
+
+export const newNode =
```
lukekarrys:
```suggestion
const newNode =
```

src/graph/test/modifiers.ts
```diff
@@ -0,0 +1,1097 @@
...
+  b --> d
+  d --> e
+  d --> f:::optional
+
+  classDef dev fill:palegreen
+  classDef optional fill:cornsilk
+```
+*/
+
+export const getSimpleGraph = (): GraphLike => {
```
lukekarrys:
```suggestion
const getSimpleGraph = (): GraphLike => {
```

src/graph/test/modifiers.ts
```diff
@@ -0,0 +1,1097 @@
...
+
+/* Returns a graph that looks like:
+```mermaid
+flowchart TD
+  root --> a
+  a --> b
+  b --> a
+```
+*/
+export const getCycleGraph = (): GraphLike => {
```
lukekarrys:
```suggestion
const getCycleGraph = (): GraphLike => {
```

src/graph/test/modifiers.ts
```diff
@@ -0,0 +1,1097 @@
...
+  root --> a
+  root --> b
+  root --> c
+  b --> a
+  b --> d
+  b --> e
+  a --> f
+```
+*/
+export const getMultiWorkspaceGraph = (): GraphLike => {
```
lukekarrys:
```suggestion
const getMultiWorkspaceGraph = (): GraphLike => {
```

src/graph/src/modifiers.ts
```diff
@@ -0,0 +1,527 @@
...
+ * Throw if the provided value is not a valid {@link GraphModiferConfig}
+ */
+export const assertGraphModifiersConfig: (
+  conf: unknown,
+  path?: string,
+) => asserts conf is GraphModiferConfig = (
+  conf: unknown,
+  path?: string,
+) => {
+  if (!conf || typeof conf !== 'object' || Array.isArray(conf)) {
```
lukekarrys: We have an `isObject` method in `@vltpkg/types` that could be used here.

lukekarrys (approved):
LGTM. Left a few suggestions/ideas but 👍🏼 on the direction and API.

src/graph/src/modifiers.ts
```diff
@@ -0,0 +1,518 @@
...
+    this.load(options)
+  }
+
+  /**
+   * Load the modifiers definitions from vlt.json,
+   * converting the result into a GraphModifierConfigObject
+   */
+  get config(): GraphModifierConfigObject {
+    if (this.#config) return this.#config
+    this.#config = asGraphModifiersConfig(
```
isaacs, line 255: You shouldn't need the "asGraphModifiersConfig" here, the `load` function ensures that the return type is either the specified type or `undefined`.

isaacs: Oh, I see, it's also switching it into a canonical type. In that case, I'd drop the assert() call in `asGraphModifiersConfig`, or name it something different so that it's clear that it's converting it from `GraphModifiersConfig` into a canonical subtype.

src/graph/test/modifiers.ts
```diff
@@ -0,0 +1,1062 @@
...
+    edgesIn: new Set(),
+    edgesOut: new Map(),
+    importer: false,
+    mainImporter: false,
+    graph,
+    id: joinDepIDTuple(['registry', '', `${name}@${version}`]),
+    name,
+    version,
+    location:
+      'node_modules/.vlt/;;${name}@${version}/node_modules/${name}',
```
isaacs, line 55: This should be `node_modules/.vlt/${joinDepIdTuple(['registry', '', name + '@' + version])}/node_modules/${name}`.

We don't use `;` anymore, because it breaks node-gyp.

PR #826, opened 2025-06-04
gui: dependency sidebar empty states

ejkorol:
Screenshot 2025-06-03 at 22 35 24

https://github.com/user-attachments/assets/ccaec0b1-004f-421c-9c93-91cc097fab29

This PR introduces new empty states for the dependency sidebar when there are no dependencies present, and an empty state for when filters are applied but no results are found.

### Other Changes
- simplifies the animation logic for the `add-dependency`'s trigger
- bugfixes the `number-flow` on the `dependency-sidebar` header when adding new dependencies
- changes the `filteredDependencies` default to `dependencies` instead of an empty array to avoid the empty state flashing

### DIFF
--- a/src/gui/src/components/explorer-grid/dependency-sidebar/filter.tsx
+++ b/src/gui/src/components/explorer-grid/dependency-sidebar/filter.tsx
@@ -6,6 +6,7 @@ import {
   ListFilter,
   CircleDot,
   Search,
+  RotateCcw,
 } from 'lucide-react'
 import { Button } from '@/components/ui/button.tsx'
 import { Input } from '@/components/ui/input.tsx'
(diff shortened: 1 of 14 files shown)


src/gui/src/components/explorer-grid/dependency-sidebar/filter.tsx
```diff
@@ -409,3 +416,99 @@ export const FilterList = () => {
...
+                  <div className="flex h-full items-center justify-between px-4">
+                    <div className="flex items-center gap-3">
+                      <div className="flex h-6 items-center justify-center rounded border-[1px] border-muted-foreground/20 bg-muted-foreground/20 px-2 font-mono text-xxs tabular-nums text-muted-foreground/60">
+                        ^19.1.0
+                      </div>
+                      <div className="text-xs font-medium text-muted-foreground/60">
+                        react-dom
+                      </div>
+                    </div>
+                    <div className="font-mono text-xxs tabular-nums text-muted-foreground/60">
```
ruyadorno: should we pull from actual dependencies here instead?

ruyadorno (commented):
I think it might be a better idea to use data from actual dependencies in the no results view for the filters.

PR #829, opened 2025-06-04
graph: rename modifier type cast helper

ruyadorno:


ruyadorno:
lol also caught a typo in that type name, let me fix that

PR #830, opened 2025-06-04
dep-id: add support to extra param in getId method

ruyadorno:
Makes it simple to get an id that uses an `extra` parameter when using the `getId` method.


src/dep-id/src/index.ts
```diff
@@ -256,7 +256,11 @@ const omitDefReg = (s?: string): string =>
  * in the manifest, registry ID types will use the name or bareSpec from the
  * specifier, so at least there's something to use later.
  */
-export const getTuple = (spec: Spec, mani: Manifest): DepIDTuple => {
+export const getTuple = (
+  spec: Spec,
+  mani: Manifest,
+  extra?: string,
+): DepIDTuple => {
```
isaacs, line 271: A lot of code can be streamlined here by just _always_ returning `..., extra]`, and treating `extra=undefined` as equivalent to it not being set.

src/dep-id/src/index.ts
```diff
@@ -96,11 +96,23 @@ export const joinDepIDTuple = (list: DepIDTuple): DepID => {
...
+      .replaceAll('%2F', '§')
+  : s
+
+const decode = (s?: string) =>
+  s ?
+    decodeURIComponent(
+      s
+        .replaceAll('@', '%40')
+        .replaceAll('§', '%2f')
+        .replaceAll('§', '%2F'),
```
lukekarrys: The last two replacements are both for the same character right? So for decoding we should just pick which one of `%2f` and `%2F` we would prefer be in the result.

lukekarrys (approved):
LGTM. One minor question/suggestion.

src/dep-id/src/index.ts
ruyadorno: good catch!

isaacs (approved):

PR #836, opened 2025-06-06
graph: add graph modifier try dependencies helper

ruyadorno:
Add a new `GraphModifier.tryDependencies` helper that takes a few of the repetitive boilerplate from the graph traversal into the modifiers helper module.

This also changes the `GraphModifier.tryNewDependency` method so that it only returns a single matching active modifier entry, which entry to return is defined by first selecting from any breadcrumb that is already complete and then using specificity to define which query should be prioritized.

### DIFF
--- a/src/graph/src/modifiers.ts
+++ b/src/graph/src/modifiers.ts
@@ -1,4 +1,7 @@
-import { parseBreadcrumb } from '@vltpkg/dss-breadcrumb'
+import {
+  parseBreadcrumb,
+  specificitySort,
+} from '@vltpkg/dss-breadcrumb'
 import { error } from '@vltpkg/error-cause'
 import { Spec } from '@vltpkg/spec'
 import { asManifest, assertRecordStringString } from '@vltpkg/types'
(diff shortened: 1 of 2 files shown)


src/graph/src/modifiers.ts
```diff
@@ -303,9 +317,56 @@ export class GraphModifier {
...
+    // it will return undefined as no entry is found in the `all` map
+    return !completeEntries.length ?
+        all.get(
+          specificitySort(arr.map(i => i.modifier.breadcrumb))[0],
+        )
+      : all.get(
+          specificitySort(
+            completeEntries.map(i => i.modifier.breadcrumb),
+          )[0],
+        )
```
lukekarrys:
```suggestion
    const entries = completeEntries.length ? completeEntries : arr
    return all.get(
      specificitySort(entries.map(i => i.modifier.breadcrumb))[0],
    )
```

Is this equivalent? If so, I find this a little more readable since the branching made me think more was going to be differentiated instead of only switching `arr` vs `completeEntries`

ruyadorno: good call! I think so!

PR #854, opened 2025-06-12
gui: less verbose query from search results

lukekarrys:
Fixes https://github.com/vltpkg/vltpkg/issues/871

### DIFF
--- a/src/gui/src/components/explorer-grid/results/result-item.tsx
+++ b/src/gui/src/components/explorer-grid/results/result-item.tsx
@@ -70,8 +70,9 @@ const onResultItemClick =
     e.preventDefault()
     if (!item.to) return
     let newQuery = ''
-    const useVersion = false as boolean
     if (item.stacked) {
+      // TODO: useVersion not needed for stacked items
+      const useVersion = false as boolean
       const name = item.to.name ? `#${item.to.name}` : ''
       const version =
         useVersion && item.to.version ? `:v(${item.to.version})` : ''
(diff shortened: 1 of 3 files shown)


lukekarrys:
Discussed with @ruyadorno and this needs some work to still allow for search results that need to be differentiated with `:v()` to properly select an item.

lukekarrys:
@ruyadorno this is ready for review again

src/gui/src/components/explorer-grid/results/result-item.tsx
```diff
@@ -70,27 +70,33 @@ const onResultItemClick =
...
       if (item.to.importer && !item.from) {
-        newQuery = `:project[name="${item.to.name}"]`
+        newQuery = `:project#${item.to.name}`
       } else if (item.from) {
-        const fromName = `[name="${item.from.name}"]`
+        // use version on the parent node if there are multiple nodes in the graph with the same name
+        const useVersion =
+          [...item.from.graph.nodes.values()].filter(
+            n => n.name === item.from?.name,
+          ).length > 1
```
ruyadorno, line 91: I noticed that the graph has this `nodesByName` property / data structure that might be particularly useful here:

```suggestion
        const nodesFound = item.from.graph.nodesByName.get(item.from?.name)
        const useVersion = nodesFound && nodesFound.size > 1
```

lukekarrys: saving this for a future optimization because `nodesByName` is not implemented on the gui graph, only the client graph.

PR #856, opened 2025-06-12
gui: Refactor useEffect to useMemo

lukekarrys:
I asked Cursor to take a look at our `useEffect` calls and see if there are any that would be better as `useMemo`. From [the docs](https://react.dev/learn/you-might-not-need-an-effect) on this:

> You don’t need Effects to transform data for rendering. For example, let’s say you want to filter a list before displaying it. You might feel tempted to write an Effect that updates a state variable when the list changes. However, this is inefficient. When you update the state, React will first call your component functions to calculate what should be on the screen. Then React will [“commit”](https://react.dev/learn/render-and-commit) these changes to the DOM, updating the screen. Then React will run your Effects. If your Effect also immediately updates the state, this restarts the whole process from scratch! To avoid the unnecessary render passes, transform all the data at the top level of your components. That code will automatically re-run whenever your props or state change.

This PR is a good start but it also found some more that are trickier to refactor because they involve lifting up other state or syncing state with the URL which should be handled with React Router. I'm going to create some follow-up issues and see if I can get Cursor to tackle those as well.

---

For transparency, here's the initial prompt I gave Cursor. I followed up with my own commits after because there was some small cleanup (that I probably could've prompted again for) but also because it seems necessary to have a non-bot user commit to get CI to run (I'm trying to figure out that limitation).

> @/gui Look through all the source code in the GUI and find any `useEffect` calls that would be better written as `useMemo` or other more idiomatic React usage.

> Read all the docs at @https://react.dev/learn/you-might-not-need-an-effect to get a good idea of the types of usage that can be refactored.


lukekarrys:
I manually checked all this functionality and agree it would be better if we had coverage for this code beyond unit tests. But I think the best way to do that would be browser integration tests with something like playwright, which is outside the scope of this PR.

PR #859, opened 2025-06-13
chore: add monorepo structure cursor rule

ruyadorno:


.cursor/rules/monorepo-structure.mdc
```diff
@@ -0,0 +1,128 @@
+---
+description: Details on how to work in the monorepo structure
+globs: 
+alwaysApply: false
```
lukekarrys: Maybe this should be `true`? For a lot of requests I'm having to tell Cursor agents how to run tests the right way.

ruyadorno: right, this one seems fundamental enough that I'd like it to apply to all my operations

```suggestion
alwaysApply: true
```

PR #860, opened 2025-06-13
gui: suggested queries

ejkorol:
Screenshot 2025-06-13 at 14 24 25

This PR creates an `overview-sidebar` component, and cleans up the 'left' side of the 'selected-item' view, and adds a `suggested-queries` component to render when the user is viewing a selected item that contains dependencies.

### Changes
- **refactors the `selected-item`'s left side**
- **creates `suggested-queries` component**
- **adds `suggested-queries` into `overview-sidebar/index.tsx`**
- **updates `selected-item`**

### DIFF
--- a/src/gui/src/components/explorer-grid/overview-sidebar/suggested-queries.tsx
+++ b/src/gui/src/components/explorer-grid/overview-sidebar/suggested-queries.tsx
@@ -1,0 +1,130 @@
+import { useEffect, useState } from 'react'
+import { ChevronRight, Search, CornerDownRight } from 'lucide-react'
+import { Button } from '@/components/ui/button.tsx'
+import { Query } from '@vltpkg/query'
+import { QueryToken } from '@/components/query-bar/query-token.tsx'
+import { useGraphStore } from '@/state/index.ts'
+import { AnimatePresence, motion } from 'framer-motion'
+
+interface SuggestedQuery {
+  name: string
+  description: string
+  query: string
+}
+
+export const queries: SuggestedQuery[] = [
+  {
+    name: 'Outdated',
+    description: 'Find packages that are outdated',
+    query: ':outdated(major)',
+  },
+  {
+    name: 'Unmaintained',
+    query: ':unmaintained',
+    description: 'Find packages that are unmaintained',
+  },
+  {
+    name: 'System access',
+    query: ':is(:fs, :env)',
+    description:
+      'Find packages with access to filesystem or environment variables',
+  },
+]
+
+export const SuggestedQueries = () => {
+  const [expanded, setExpanded] = useState<boolean | null>(null)
+
+  useEffect(() => {
+    const stored = localStorage.getItem('suggestedQueriesExpanded')
+    setExpanded(stored === 'true')
+  }, [])
+
+  const query = useGraphStore(state => state.query)
+  const updateQuery = useGraphStore(state => state.updateQuery)
+
+  const toggleExpanded = () => {
+    setExpanded(prev => {
+      const newValue = !prev
+      localStorage.setItem(
+        'suggestedQueriesExpanded',
+        String(newValue),
+      )
+      return newValue
+    })
+  }
+
+  const runQuery = (suggestedQuery: string) =>
+    updateQuery(`${query} ${suggestedQuery}`)
+
+  if (expanded === null) return null
+
+  return (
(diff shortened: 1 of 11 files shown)


src/gui/src/components/explorer-grid/overview-sidebar/suggested-queries.tsx
```diff
@@ -0,0 +1,130 @@
...
+  {
+    name: 'Unmaintained',
+    query: ':unmaintained',
+    description: 'Find packages that are unmaintained',
+  },
+  {
+    name: 'System access',
+    query: ':has(:fs) :has(:env)',
+    description:
+      'Find packages with access to filesystem and environment variables',
```
ruyadorno: I think the correct description for this query would be:

```suggestion
      'Find packages that contains dependencies with access to filesystem or environment variables',
```

if you want a query that actually behaves like the description then we'd need to change `:has()` by `:is()`, e.g:

```
:is(:fs):is(:env)
```

src/gui/src/components/explorer-grid/overview-sidebar/suggested-queries.tsx
```diff
@@ -0,0 +1,130 @@
...
+    query: ':outdated(major)',
+  },
+  {
+    name: 'Unmaintained',
+    query: ':unmaintained',
+    description: 'Find packages that are unmaintained',
+  },
+  {
+    name: 'System access',
+    query: ':has(:fs) :has(:env)',
```
ruyadorno: I think the space here is a mistake?

```suggestion
    query: ':has(:fs):has(:env)',
```

src/gui/test/components/explorer-grid/overview-sidebar/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+vi.mock('@/components/explorer-grid/header.tsx', () => ({
+  GridHeader: 'gui-grid-header',
+}))
+
+vi.mock(
+  '@/components/explorer-grid/overview-sidebar/suggested-queries.tsx',
+  () => ({
+    SuggestedQueries: 'gui-suggested-queries',
+  }),
+)
```
ruyadorno, line 22: we want to maybe mock the SideItem here?

`import { SideItem } from '@/components/explorer-grid/side-item.tsx`

why is it mocked in some tests but not others? 🤔

src/gui/src/components/explorer-grid/overview-sidebar/suggested-queries.tsx
darcyclarke: I believe what you want is `:is(:fs, :env)`... `:is(...):is(...)` seems wrong/redundant. You can also just us a `,` (ie. `:fs, :env`) although that isn't a "forgiving selector" like `:is(...,...)`.

src/gui/src/components/explorer-grid/overview-sidebar/suggested-queries.tsx
darcyclarke: Agreed that the space seems like it's a mistake... that said, `:is()` should be able to nest in `:has()` to remove the multiple `:has` statements (ie. `:has(:is(:fs,:env))`) that said, you should likely be looking up elements themselves & not elements with children that match this criteria.

ie. `:is(:fs, :env)` is likely the more appropriate selector which will match parents & children that have file system or environment variable access.

src/gui/test/components/explorer-grid/overview-sidebar/index.tsx
ejkorol: Good catch, I initially thought to mock `SideItem` only for certain tests, so that I could test the actual behaviour of the `onSelect`, but actually moving this test directly into `SideItem` makes more sense

src/gui/src/components/explorer-grid/overview-sidebar/suggested-queries.tsx
```diff
@@ -25,9 +25,9 @@ export const queries: SuggestedQuery[] = [
   },
   {
     name: 'System access',
-    query: ':has(:fs) :has(:env)',
+    query: ':is(:fs, :env)',
     description:
-      'Find packages with access to filesystem and environment variables',
+      'Find packages that contain dependencies with access to filesystem or environment variables',
```
ruyadorno: now that the query changed the original text was almost correct 😅 
```suggestion
      'Find packages with access to filesystem or environment variables',
```

ruyadorno (approved):

Issue #801, opened 2025-05-28
[BUG] tap fails with react-reconciler error


lukekarrys:
```
╰❯ vlx tap
/Users/lukekarrys/projects/vltpkg/vltpkg/node_modules/.vlt/··react-reconciler@0.29.2/node_modules/react-reconciler/cjs/react-reconciler.development.js:4279
  throw new Error("Objects are not valid as a React child (found: " + (childString === '[object Object]' ? 'object with keys {' + Object.keys(newChild).join(', ') + '}' : childString) + "). " + 'If you meant to render a collection of children, use an array ' + 'instead.');
        ^

Error: Objects are not valid as a React child (found: object with keys {$$typeof, type, key, props, _owner, _store}). If you meant to render a collection of children, use an array instead.
    at throwOnInvalidObjectType (/Users/lukekarrys/projects/vltpkg/vltpkg/node_modules/.vlt/··react-reconciler@0.29.2/node_modules/react-reconciler/cjs/react-reconciler.development.js:4279:9)
    at reconcileChildFibers (/Users/lukekarrys/projects/vltpkg/vltpkg/node_modules/.vlt/··react-reconciler@0.29.2/node_modules/react-reconciler/cjs/react-reconciler.development.js:5220:7)
    at reconcileChildren (/Users/lukekarrys/projects/vltpkg/vltpkg/node_modules/.vlt/··react-reconciler@0.29.2/node_modules/react-reconciler/cjs/react-reconciler.development.js:10366:28)
    at updateHostRoot (/Users/lukekarrys/projects/vltpkg/vltpkg/node_modules/.vlt/··react-reconciler@0.29.2/node_modules/react-reconciler/cjs/react-reconciler.development.js:11085:5)
    at beginWork (/Users/lukekarrys/projects/vltpkg/vltpkg/node_modules/.vlt/··react-reconciler@0.29.2/node_modules/react-reconciler/cjs/react-reconciler.development.js:12827:14)
    at beginWork$1 (/Users/lukekarrys/projects/vltpkg/vltpkg/node_modules/.vlt/··react-reconciler@0.29.2/node_modules/react-reconciler/cjs/react-reconciler.development.js:19608:14)
    at performUnitOfWork (/Users/lukekarrys/projects/vltpkg/vltpkg/node_modules/.vlt/··react-reconciler@0.29.2/node_modules/react-reconciler/cjs/react-reconciler.development.js:18742:12)
[...]

PR #866, opened 2025-06-16
fix: resolution issues (modifiers/duplicate nodes etc.)

darcyclarke:
### Addressing
- [x] ensure dep type is correctly calculated during resolution
- [x] duplicate packages are no longer installed for modifiers
- [ ] modifiers shouldn't fork deps if not necessary (ie. lean into existing resolution algo)
- [ ] duplicate packages are no longer installed for the default registry (ie. `··` & `·npm·` were being treated as different depIDs which shouldn't have been possible)
- [ ] nodes & artifacts are detached

### DIFF
--- a/src/graph/src/graph.ts
+++ b/src/graph/src/graph.ts
@@ -342,16 +342,15 @@ export class Graph implements GraphLike {
         depType === 'peerOptional',
     }
 
+    // Don't include modifier information in regular dependency node IDs
+    // Modifiers are stored as a property on the node, not in the ID
+    // This ensures that normal dependencies can be properly resolved and cached
+    const depId = id || (manifest && getId(spec, manifest))
     const saveType = resolveSaveType(
       fromNode,
       spec.final.name,
       depType,
     )
-
-    // Don't include modifier information in regular dependency node IDs
-    // Modifiers are stored as a property on the node, not in the ID
-    // This ensures that normal dependencies can be properly resolved and cached
-    const depId = id || (manifest && getId(spec, manifest))
 
     /* c8 ignore start - should not be possible */
     if (!depId) {


src/graph/src/graph.ts
```diff
@@ -370,8 +378,13 @@ export class Graph implements GraphLike {
     toNode.registry = spec.registry
     toNode.dev = flags.dev
     toNode.optional = flags.optional
-    toNode.modifier = extra
-    this.addEdge(depType, spec, fromNode, toNode)
+
+    // Set the modifier if provided via extra parameter
+    if (extra) {
+      toNode.modifier = extra
+    }
```
lukekarrys, line 385: Not sure if this addition is relevant to this change. iirc @ruyadorno opted to always apply the modifier to the node even if it was `undefined`

src/graph/src/graph.ts
```diff
@@ -241,7 +241,7 @@ export class Graph implements GraphLike {
   }
 
   /**
-   * Find an existing node to satisfy a dependency
+   * Find an existing node in the graph that satisfies a given {@link Spec}.
```
lukekarrys:
```suggestion
   * Find an existing node to satisfy a dependency
```

src/graph/src/graph.ts
```diff
@@ -253,6 +253,7 @@ export class Graph implements GraphLike {
     if (cached) return cached
     const nbn = this.nodesByName.get(f.name)
     if (!nbn) return undefined
+
```
lukekarrys:
```suggestion
```

src/graph/src/graph.ts
```diff
@@ -269,6 +270,7 @@ export class Graph implements GraphLike {
         return node
       }
     }
+    return undefined
```
lukekarrys:
```suggestion
```

src/graph/src/graph.ts
```diff
@@ -342,7 +344,13 @@ export class Graph implements GraphLike {
         depType === 'peerOptional',
     }
 
-    const depId = id || (manifest && getId(spec, manifest, extra))
+    const f = spec.final
+    const saveType = resolveSaveType(fromNode, f.name, depType)
```
lukekarrys:
```suggestion
    const saveType = resolveSaveType(
      fromNode,
      spec.final.name,
      depType,
    )
```

src/graph/src/graph.ts
```diff
@@ -342,7 +342,16 @@ export class Graph implements GraphLike {
...
+    const saveType = resolveSaveType(
+      fromNode,
+      spec.final.name,
+      depType,
+    )
+
+    // Don't include modifier information in regular dependency node IDs
+    // Modifiers are stored as a property on the node, not in the ID
+    // This ensures that normal dependencies can be properly resolved and cached
+    const depId = id || (manifest && getId(spec, manifest))
```
lukekarrys:
```suggestion
    // Don't include modifier information in regular dependency node IDs
    // Modifiers are stored as a property on the node, not in the ID
    // This ensures that normal dependencies can be properly resolved and cached
    const depId = id || (manifest && getId(spec, manifest))
    const saveType = resolveSaveType(
      fromNode,
      spec.final.name,
      depType,
    )
```

src/graph/src/graph.ts
lukekarrys: Testing more it is relevant in that it changes the error from #801 to different error. But would like @ruyadorno's opinion on if it changes anything with modifiers.

src/graph/src/graph.ts
```diff
@@ -342,7 +342,15 @@ export class Graph implements GraphLike {
         depType === 'peerOptional',
     }
 
-    const depId = id || (manifest && getId(spec, manifest, extra))
+    // Don't include modifier information in regular dependency node IDs
+    // Modifiers are stored as a property on the node, not in the ID
+    // This ensures that normal dependencies can be properly resolved and cached
+    const depId = id || (manifest && getId(spec, manifest))
```
lukekarrys, line 348: Same question as below applies here. My understanding is that `extra` was being used to make sure the `depId` would match a modified dep and that would need to be done in addition to assigning it to `toNode.modifer`.

lukekarrys: But my simple tests of `vlt i react@19 && echo '{"modifiers": { "#react": "18.0.0"}}' > vlt.json && vlt i` does still work with this branch.

ruyadorno (commented):
Synced with @darcyclarke offline, let's repurpose this PR and try to document all the specific issues found instead.

PR #880, opened 2025-06-18
dev: add cursor rule about running single file tests

lukekarrys:


.cursor/rules/monorepo-structure.mdc
```diff
@@ -115,6 +115,12 @@ actions:
       2. Each workspace has its own `package.json` and dependencies
       3. Unit tests are located in each workspace's `test/` folder
 
+      **Running Tests:**
+      - To run tests for a single file, use: `pnpm --filter ./src/<workspace> test -- ./test/<test-file>`
+      - Example: `pnpm --filter ./src/cli-sdk test -- ./test/load-command.ts`
+      - Tests require 100% coverage to pass - the command will exit with status 1 if coverage is incomplete
+      - Test results will show coverage metrics for statements, branches, functions, and lines
+
```
ruyadorno, line 123: What about having it in [`.cursor/rules/code-validation-workflow.mdc`](https://github.com/vltpkg/vltpkg/blob/main/.cursor/rules/code-validation-workflow.mdc) instead? 🤔

lukekarrys: I missed that file and I see it already has instructions for running single file tests. So maybe this isn't needed here?

For this PR I gave Cursor a bunch of instructions and corrections on how to edit a single file and then run tests for that file, and then I asked it to update any relevant `.cursor/rules` files.

ruyadorno: oooohhh I see, cool cool

PR #890, opened 2025-06-22
graph: refactor modifier state changes

ruyadorno:
Changed the logic for building a new ideal graph to make sure a new graph is completely rebuild in case modifiers were changed since a previous install, instead of starting from a lockfile/actual loaded graph.

To have a reference on what modifiers were used before, modifiers are now properly stored in the lockfile options section along with having the config `vlt.json` file copied to the `node_modules/.vlt` store folder.

A few notable internal changes are included:
- Implement `skipLoadingNodesOnModifiersChange` option to conditionally load dependencies in both `actual` and `lockfile` load methods
- Refactor lockfile format to store modifiers in options field
- Add `modifier` and `registry` fields to `NodeLike` interface
- Update dependency parsing logic to handle modifiers in `DepID` `extra` parameters
- Add `removeEdgeResolution` method to the `Graph` class
- Added new `getRawDependencies` and `getDependencies` functions to the `dependencies` module
- Add extensive test coverage for new modifier functionality and type validation


ruyadorno:
after syncing with @darcyclarke earlier this week I realized we're going to need to keep state of the currently applied modifiers attached to both the virtual and actual graphs so that we can choose to ignore those when rebuilding the ideal graph to apply changed modifiers.

src/graph/src/ideal/build-ideal-from-starting-graph.ts
```diff
@@ -59,7 +56,7 @@ export const buildIdealFromStartingGraph = async (
   await addNodes(options)
 
   // check nodes for modifiers
-  await checkNodes({ check: importerSpecs.check, ...options })
+  //await checkNodes({ check: importerSpecs.check, ...options })
```
lukekarrys: Should this be removed?

src/graph/src/reify/index.ts
```diff
@@ -124,7 +127,7 @@ const reify_ = async (
   lockfile.save(options)
 
   // if we had to change the actual graph along the way,
-  // make sure we do not leave behind any unreachable nodes
+  // make sure we do njkot leave behind any unreachable nodes
```
lukekarrys:
```suggestion
  // make sure we do not leave behind any unreachable nodes
```

src/graph/src/actual/load.ts
```diff
@@ -380,29 +371,55 @@ export const load = (options: LoadOptions): Graph => {
...
-    depsFound.set(
-      importer,
-      scurry.cwd.resolve(`${importer.location}/node_modules`),
+  // retrieve the configuration object from the store
+  let storeConfig: StoreConfigObject = { modifiers: undefined }
+  try {
+    storeConfig = asStoreConfigObject(
+      JSON.parse(
+        readFileSync(
+          scurry.resolve('node_modules/.vlt/vlt.json'),
```
lukekarrys, line 381: Is this different than the `options` we store in hidden lockfile (`node_modules/.vlt-lock.json`)? We already store the `catalog` entries there:

```
$ cat vlt.json
{
  "catalog": {
    "typescript": "^5.0.0"
  },
  "catalogs": {
    "lint": {
      "eslint": "^9.0.0"
    }
  }
}
                                                                                                                    
$ cat node_modules/.vlt-lock.json | jq .options
{
  "catalog": {
    "typescript": "^5.0.0"
  },
  "catalogs": {
    "lint": {
      "eslint": "^9.0.0"
    }
  }
}
```

lukekarrys:
This PR is failing because in CI the workflow running `vlt install` is installing `prettier@3.6.0` (which was published ~14 hours ago) even though the lockfile contains `prettier@3.5.3`. I'm not able to reproduce this locally (tried deleting cache and reinstalling with `vlt i`), but I don't think it's worth investing and holding up this PR. But this could be a bug in `vlt` that we'll want to investigate.

So this PR can be merged even though it's red and I will make a follow-up to update prettier  which will fix CI.

src/graph/src/ideal/build-ideal-from-starting-graph.ts
ruyadorno:
```suggestion
```

PR #899, opened 2025-06-24
cli-sdk: add version command

lukekarrys:


src/cli-sdk/src/commands/version.ts
```diff
@@ -0,0 +1,230 @@
...
+    spawn_(args, { cwd, ...opts })
+
+  if (!increment) {
+    throw error('Version increment argument is required', {
+      code: 'EUSAGE',
+      validOptions: versionIncrements,
+    })
+  }
+
+  const manifest = conf.options.packageJson.read(cwd)
```
ruyadorno: I just added a very handy method in https://github.com/vltpkg/vltpkg/pull/940 that we can potentially use here instead:
```suggestion
  const manifest = conf.options.packageJson.find()
```

That's going to walk up the directory tree and operate on the first found `package.json` file, which is going to allow this command to also work from nested directories

darcyclarke (approved):

PR #908, opened 2025-06-25
gui: fix create query tooltip

ruyadorno:
The query tooltip for the saved category color was being cropped by the top level background.

Screenshot 2025-06-24 at 9 33 54 PM


ruyadorno:
Want you to double check here @ejkorol since I couldn't reproduce it after I fixed it.

ejkorol:
> Want you to double check here @ejkorol since I couldn't reproduce it after I fixed it.

Yes this is the fix!

PR #925, opened 2025-06-25
gui: fix search result root nav

ruyadorno:
Fix clicking on search results when the current query starts from a `:root` reference.

### DIFF
--- a/src/gui/test/lib/update-result-item.test.ts
+++ b/src/gui/test/lib/update-result-item.test.ts
@@ -1,0 +1,765 @@
+import { describe, it, expect, vi } from 'vitest'
+import { updateResultItem } from '@/lib/update-result-item.ts'
+import type { GridItemData } from '@/components/explorer-grid/types.ts'
+import type { QueryResponseNode } from '@vltpkg/query'
+import { joinDepIDTuple } from '@vltpkg/dep-id'
+
+// Helper function to create a mock QueryResponseNode
+const createMockNode = (
+  name: string,
+  version: string,
+  importer = false,
+  id?: string,
+): QueryResponseNode =>
+  ({
+    name,
+    version,
+    importer,
+    id: id || joinDepIDTuple(['registry', '', `${name}@${version}`]),
+    graph: {
+      nodes: new Map([
+        [
+          joinDepIDTuple(['registry', '', `${name}@${version}`]),
+          { name, version },
+        ],
+      ]),
+    } as any,
+    edgesIn: new Set(),
+    edgesOut: new Map(),
+    insights: {
+      scanned: false,
+    },
+    toJSON: () => ({}) as any,
+  }) as QueryResponseNode
+
+// Helper function to create a mock QueryResponseNode with multiple versions
+const createNodeWithMultipleVersions = (
+  name: string,
+  version: string,
+  otherVersions: string[],
+): QueryResponseNode => {
+  const nodes = new Map()
+  nodes.set(joinDepIDTuple(['registry', '', `${name}@${version}`]), {
+    name,
+    version,
+  })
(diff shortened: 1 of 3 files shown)


ruyadorno:
I can already see this is highlighting a few broken scenarios, we'll land since it already has at least one fix but we should iterate on fixing those later.

src/gui/test/lib/update-result-item.test.ts
```diff
@@ -0,0 +1,765 @@
...
+      const handler = updateResultItem({
+        item,
+        query: '',
+        updateQuery: mockUpdateQuery,
+      })
+
+      handler(mockEvent)
+
+      expect(mockEvent.preventDefault).toHaveBeenCalled()
+      expect(mockUpdateQuery).toHaveBeenCalledWith(':project#')
```
ruyadorno, line 248: ⚠️ this should never happen

src/gui/test/lib/update-result-item.test.ts
```diff
@@ -0,0 +1,765 @@
...
+        item,
+        query: ':root some-other-query',
+        updateQuery: mockUpdateQuery,
+      })
+
+      handler(mockEvent)
+
+      expect(mockEvent.preventDefault).toHaveBeenCalled()
+      expect(mockUpdateQuery).toHaveBeenCalledWith(
+        ':root #parent-package > some-other-query',
```
ruyadorno, line 374: should pb use a valid query example here instead of `some-other-query`

src/gui/test/lib/update-result-item.test.ts
```diff
@@ -0,0 +1,765 @@
...
+        item,
+        query: '',
+        updateQuery: mockUpdateQuery,
+      })
+
+      handler(mockEvent)
+
+      expect(mockEvent.preventDefault).toHaveBeenCalled()
+      expect(mockUpdateQuery).toHaveBeenCalledWith(
+        '#parent-package > ',
```
ruyadorno, line 436: this is broken

src/gui/test/lib/update-result-item.test.ts
```diff
@@ -0,0 +1,765 @@
...
+        item,
+        query: 'existing query',
+        updateQuery: mockUpdateQuery,
+      })
+
+      handler(mockEvent)
+
+      expect(mockEvent.preventDefault).toHaveBeenCalled()
+      expect(mockUpdateQuery).toHaveBeenCalledWith(
+        'existing query#simple-package',
```
ruyadorno, line 536: unclear if this is a broken state or not, again, using a valid example will help clarify

src/gui/test/lib/update-result-item.test.ts
```diff
@@ -0,0 +1,765 @@
...
+        item,
+        query: '   trimmed query   ',
+        updateQuery: mockUpdateQuery,
+      })
+
+      handler(mockEvent)
+
+      expect(mockEvent.preventDefault).toHaveBeenCalled()
+      expect(mockUpdateQuery).toHaveBeenCalledWith(
+        'trimmed query#package',
```
ruyadorno, line 588: same

src/gui/test/lib/update-result-item.test.ts
```diff
@@ -0,0 +1,765 @@
...
+      const handler = updateResultItem({
+        item,
+        query: '',
+        updateQuery: mockUpdateQuery,
+      })
+
+      handler(mockEvent)
+
+      expect(mockEvent.preventDefault).toHaveBeenCalled()
+      expect(mockUpdateQuery).toHaveBeenCalledWith('# > #child')
```
ruyadorno, line 706: broken? maybe it's an invalid fixture

PR #930, opened 2025-06-26
Add pack and publish commands

lukekarrys:


src/cli-sdk/src/commands/pack.ts
```diff
@@ -0,0 +1,87 @@
...
+      `📦 ${r.name}@${r.version}`,
+      `📄 ${r.filename}`,
+      `📁 ${r.files.length} files`,
+      ...r.files.map(f => `  - ${f}`),
+      `📊 package size: ${prettyBytes(r.size)}`,
+      `📂 unpacked size: ${prettyBytes(r.unpackedSize)}`,
+    ]
+    if (r.shasum) lines.push(`🔒 shasum: ${r.shasum}`)
+    if (r.integrity) lines.push(`🔐 integrity: ${r.integrity}`)
+    return lines.join('\n')
```
ruyadorno, line 50: Love the emojis! we need to do more of those

src/cli-sdk/src/pack-tarball.ts
ruyadorno, line 1: this should be its own workspace at `@vltpkg/pack` or `@vltpkg/pack-tarball` or something so that it's going to be simpler to integrate in the GUI backend server, it paid off for `@vltpkg/init`!

src/cli-sdk/src/pack-tarball.ts
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        // Default behavior when no files field - exclude common development files
+        const defaultExcludePatterns = [
+          /^\.?\/?\.vscode(\/|$)/,
+          /^\.?\/?\.idea(\/|$)/,
+          /^\.?\/?\.gitignore$/,
+          /^\.?\/?\.npmignore$/,
+          /^\.?\/?\.editorconfig$/,
+        ]
```
ruyadorno, line 141: talking from experience, I think we'd benefit from having these defined as separate files so that we can point at them across documentation anytime either people want to consult them or propose changes, it applies to all the other patterns too

src/cli-sdk/src/pack-tarball.ts
```diff
@@ -0,0 +1,196 @@
...
+
+        // Always include certain files
+        const alwaysIncludePatterns = [
+          /^package\.json$/,
+          /^README(\..*)?$/i,
+          /^CHANGELOG(\..*)?$/i,
+          /^HISTORY(\..*)?$/i,
+          /^LICENSE(\..*)?$/i,
+          /^LICENCE(\..*)?$/i,
+        ]
```
ruyadorno, line 85: question: on these and the other patterns, are these lists 1:1 to npm?

src/cli-sdk/src/commands/publish.ts
```diff
@@ -0,0 +1,175 @@
...
+    ]
+    if (r.shasum) lines.push(`🔒 shasum: ${r.shasum}`)
+    if (r.integrity) lines.push(`🔐 integrity: ${r.integrity}`)
+
+    return lines.join('\n')
+  },
+  json: r => r,
+} as const satisfies Views<CommandResult>
+
+export const command: CommandFn<CommandResult> = async conf => {
```
ruyadorno, line 71: I believe this entire method should be defined as its own package (e.g: `@vltpkg/publish`), similar to my comment on `src/cli-sdk/src/pack-tarball.ts` so that it can easily be reusable in other places like the GUI server

ruyadorno (approved):
I love the direction here ❤️ 

Some optional feedback also left in comments:
I would love to have the logic be already available as separate workspaces / packages so that it reduces the effort required to bring these features to the GUI later and one other item of note is having all the pattern definitions as separate files is an easy win in terms of maintainability IMO.

While I could not test those, I think it's also essential that we handle `workspaces:` and `catalog:` specs properly. If they're not being handled atm then it needs to be added in a quick follow up.

lukekarrys:
@ruyadorno Great feedback. I agree that pack and publish could be separate workspaces which would help with documentation and discoverability for the list of add/exclude files. I will add tasks to the follow-up work I have for this feature set.

I will also add TODOs for the `catalog:` and `workspace:` specs which are not being handled currently.

PR #947, opened 2025-06-28
gui: update contributor list

ejkorol:
### Overview
The previous contributor list in the overview tab was visually distracting and chaotic, this PR resolves this issue by creating a new hidden `tabs-contributors` view within the selected item, that is accessible by clicking 'see all contributors' button, and provides a cleaner and more organized view of contributors.

Additionally, this PR aggregates the `contributors` and `maintainers` not from the `manifest` but from the `packument`

| overview tab | contributor tab |
| --- | --- |
| Screenshot 2025-06-28 at 03 00 29 | Screenshot 2025-06-28 at 03 00 36 |

### Features
- **feat: creates `tabs-contributors`**
- **removes old contributor components from `tabs-overview`**
- **feat: creates new routes and tab types**

### DIFF
--- a/src/gui/src/lib/external-info.ts
+++ b/src/gui/src/lib/external-info.ts
@@ -1,12 +1,16 @@
 import { normalizeManifest } from '@vltpkg/types'
 import { compare, gt, prerelease } from '@vltpkg/semver'
-import { isRecord } from '@/utils/typeguards.ts'
 import {
   getRepoOrigin,
   getRepositoryApiUrl,
 } from '@/utils/get-repo-url.ts'
 import type { Spec } from '@vltpkg/spec/browser'
-import type { Repository, Manifest, Packument } from '@vltpkg/types'
+import type {
+  Repository,
+  Manifest,
+  Packument,
+  NormalizedContributor,
+} from '@vltpkg/types'
 
 export type Semver = `${number}.${number}.${number}`
 
@@ -71,9 +75,7 @@ export type Version = {
   gitHead?: string
 }
 
-export type Contributor = {
-  name?: string
-  email?: string
+export type Contributor = NormalizedContributor & {
   avatar?: string
 }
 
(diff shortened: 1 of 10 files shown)


ruyadorno:
The current state of navigation is a bit confusing, is `/<encode>/contributors` supposed to be:

1. a top-level tab (in which case it's missing from the layout atm) **OR**
2. is it supposed to be a subnav of the "Overview" tab (in which case the route should be `/<encode>/overview/contributors` and should probably have a back button) ???

src/gui/src/lib/external-info.ts
```diff
@@ -455,6 +463,7 @@ export async function* fetchDetails(
             })
 
           const resolvedVersions = await Promise.all(versions)
+          const contributors = await getAllContributors(packu)
```
ruyadorno: I believe the `contributors` info should be added as part of the `.map(async ([version, mani]) => {` function above, once this PR is rebased you'll notice that the manifest is already normalized at that point, so you should be able to just use the information for that version similar to all the returned values there such as `version`, `publishedData`, `unpackedSize`, etc

src/gui/src/lib/external-info.ts
```diff
@@ -131,6 +132,41 @@ export const readRepository = (
...
+    deduped.map(async person => ({
+      name: person.name.trim(),
+      email: person.email?.trim(),
+      avatar: await retrieveAvatar(person.email || ''),
+    })),
+  )
+
+  return contributors
+}
+
```
ruyadorno, line 150: I'm not sure what is the final goal (is the `packu.contributors` and `packu.maintainers` info really necessary, desired here? at this point the per-version `contributors` info should already be available) but either way it should reuse the `normalizeContributors()` method from `@vltpkg/types` if you need to normalize these values here

ejkorol:
> The current state of navigation is a bit confusing, is `/<encode>/contributors` supposed to be:
> 
> 1. a top-level tab (in which case it's missing from the layout atm) **OR**
> 2. is it supposed to be a subnav of the "Overview" tab (in which case the route should be `/<encode>/overview/contributors` and should probably have a back button) ???

The way i've thought about it for this implementation, is that `/<encode>/contributors` is a _'hidden'_ top-level tab, It doesn't necessarily need to have it's own `tab-button` in the `tabs-list` (since it's not _really_ something that you'd always click on / want to take up the real estate on the `tabs-list`) and it's not a nested route under `/contributors` either (since it doesn't have a shared layout)

It's kind of how when you click on (for example) 'releases' on a github page, though releases doesn't have an actual button in their tab list, it still remains a kind of 'hidden' top-level tab

Though I do agree having a back button on the contributors content is good! I've added that just for a quick way to go 'backwards'

darcyclarke (approved):

ruyadorno (approved):

PR #957, opened 2025-07-01
gui: update snapshots (until next year)

lukekarrys:


ruyadorno (approved):
we should mock this value 😅

PR #960, opened 2025-07-02
graph: add edge modifier removal

ruyadorno:
Add support to a `"-"` `bareSpec` definition in the graph modifiers that enables removing a given edge from the graph.

Fixes: https://github.com/vltpkg/statusboard/issues/175


src/graph/test/ideal/append-nodes.ts
```diff
@@ -593,143 +609,120 @@ t.test(
...
-          },
-          interactiveBreadcrumb: {
-            current: breadcrumbItem,
-            next: () => true,
-            done: true,
-          },
-          originalFrom: graph.mainImporter,
-        },
-      ],
-    ])
```
ruyadorno, line 629: Replaced the "mock everything" approach with a more integrated style for this one, that will actually load modifiers from the fixture `vlt.json` file.

lukekarrys: Yes! This is the right approach.

PR #961, opened 2025-07-02
gui: feat breadcrumb navigation

ejkorol:
Screenshot 2025-07-02 at 13 39 45

### Overview
This PR introduces a new way of navigating the selected item's ancestry by adding a `crumb-nav` component. The `crumb-nav` is placed into the selected item's `item-header` and utilizes `@vltpkg/dss-breadcrumb` to parse the breadcrumbs from the current query in the `state` to render the trail.
This allows quick navigation to the parent items, making a nice quality of life improvement

### Changelog
- **feat: creates `crumb-nav` to navigate selected item ancestry**
- **feat: integrates `crumb-nav` into the `item-header`**

### DIFF
--- a/src/gui/src/components/navigation/crumb-nav.tsx
+++ b/src/gui/src/components/navigation/crumb-nav.tsx
@@ -1,0 +1,180 @@
+import React, { useMemo, useState } from 'react'
+import { useGraphStore } from '@/state/index.ts'
+import { QueryToken } from '@/components/query-bar/query-token.tsx'
+import { Query } from '@vltpkg/query'
+import { cn } from '@/lib/utils.ts'
+import { parseBreadcrumb } from '@vltpkg/dss-breadcrumb'
+import { ChevronRight, Ellipsis } from 'lucide-react'
+
+import type { GridItemData } from '@/components/explorer-grid/types.ts'
+import type { ParsedSelectorToken } from '@vltpkg/query'
+import type { State } from '@/state/types'
+
+interface CrumbNavProps {
+  className?: string
+  breadcrumbs: GridItemData['breadcrumbs']
+}
+
+interface CrumbProps {
+  query: string
+  crumbIdx: number
+  crumbLength: number
+}
+
+const MAX_BREADCRUMBS = 5
+const TRUNCATED_TAIL_COUNT = 3
+
+export const getBreadcrumbs = (
+  query: State['query'],
+): GridItemData['breadcrumbs'] | undefined => {
+  try {
+    const breadcrumb = parseBreadcrumb(query)
+
+    if (breadcrumb.single && breadcrumb.first.value === ':root') {
+      return undefined
+    }
+
+    return breadcrumb
+  } catch (_e) {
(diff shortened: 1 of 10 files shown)


src/gui/src/components/navigation/crumb-nav.tsx
```diff
@@ -0,0 +1,179 @@
...
+    const breadcrumb = parseBreadcrumb(query)
+
+    if (breadcrumb.single && breadcrumb.first.value === ':root') {
+      return undefined
+    }
+
+    return breadcrumb
+  } catch (_e) {
+    return undefined
+  }
```
ruyadorno, line 40: The support is so limited for the parsed breadcrumbs (using `@vltpkg/dss-breadcrumb`) at the moment that I'd prefer to hold on to landing this PR until we have more of the syntax there, for example rn we can't handle queries like:
- `:root > * > #foo`
- `:root >:not(:unpopular) > *:is(#ink > #ansi-styles)`
- `:root :has(:unpopular, :semver(>4)) ~ #chalk`
(to name a few examples I tried on a project)

I feel like the implementation here is going to need a few tweaks since right now `parseBreadcrumb` is not returning combinators info and is not capable of handling nested pseudo selectors

src/gui/src/components/navigation/crumb-nav.tsx
ruyadorno, line 1: [nit] I feel like the rendered items of the breadcrumb should have a different style from the style used in the query bar, I think it might be even better for them to have no color at all but if they do preserve a similar style, I'd like for it to at least to have a larger padding on each side, anyways, definitely something I'll defer to you and @darcyclarke to decide

src/gui/src/components/navigation/crumb-nav.tsx
```diff
@@ -0,0 +1,179 @@
...
+              )}
+
+            <button
+              className="inline-flex"
+              onClick={() => onCrumbClick(originalIndex)}>
+              <Crumb
+                crumbLength={parts.length}
+                crumbIdx={originalIndex}
+                query={part.value}
+              />
```
ruyadorno: I think a clickable breadcrumb item should only render its `part.name` if available (as in `#${part.name}`), otherwise if it's `*` or some other token that has no name (note: none of that is supported by the current `@vltpkg/dss-breadcrumb` implementation) it may render but not be clickable but most importantly, in case a single item that has a name, also has more pseudo selector information attached to it, then we should just drop all the extra information.

---

For example, in this case:

Screenshot 2025-07-02 at 2 46 12 PM


### It should render only:

```
#utf-8-validate > #node-gyp-build
```

ruyadorno (changes requested):

src/gui/src/components/navigation/crumb-nav.tsx
ejkorol: The syntax is pretty basic right now, but as we continue to expand what `dss-breadcrumb` allows us to parse, then more data could be potentially reflected here. What do you think?

PR #977, opened 2025-07-03
fix: normalize "contributors"

darcyclarke:
### Description
- normalize `"contributors"` with `"maintainers"` & create a managed object which adds net-new fields for the properties of what a "maintainer" historically would be (ie. "maintainers" are formally treated as a subset of "contributors" with either write access/publisher status for the associative version)
- this consolidates the understanding of "contributors" deduplicating by the associative `"email"`
- remove `deno` smoke tests (which have issues)

### DIFF
--- a/infra/build/src/variants.ts
+++ b/infra/build/src/variants.ts
@@ -5,22 +5,14 @@ import { compile } from './compile.ts'
 import { BINS } from './bins.ts'
 import type { Bin } from './bins.ts'
 
-export const VARIANTS = [
-  'Node',
-  'Deno',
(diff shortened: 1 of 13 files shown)


darcyclarke:
@lukekarrys `@`ing you for a review as I'm removing some `deno`-specific config in the infra smoke tests here since it was getting tripped up/erroring. Notably, I still need to make a couple of other changes to make sure we're doing this work in all the right places (after I got some good feedback/review from @ruyadorno).

ruyadorno (approved):

PR #983, opened 2025-07-07
gui: dashboard card design update

### LINKED ISSUE #868: [BUG] GUI: Fix Dashboard Card Design
The background logos are pretty distracting - I think we can keep the logos but maybe change them to be smaller/nested in the bottom right of the top section of the card. Open to ideas here including reintroducing the color to the logos instead of keeping them greyscale. They should definitely be smaller though & having them as a background motif was a nice idea (mine btw) but just isn't as good as it could be.

ejkorol:
Screenshot 2025-07-07 at 09 54 16

### Overview

This PR simplifies the `dashboard-item` card design. It moves the `dashboard-item` into a separate file and updates the icon position and style of the cards.

Closes #868 

### Changelog
- **feat: moves `dashboard-item` into its own component**
- **feat: simplifies dashbaord card design**

### DIFF
--- a/src/gui/src/components/dashboard-grid/dashboard-item.tsx
+++ b/src/gui/src/components/dashboard-grid/dashboard-item.tsx
@@ -1,0 +1,89 @@
+import { CardTitle } from '@/components/ui/card.tsx'
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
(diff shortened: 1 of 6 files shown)


darcyclarke (approved):
Looks great!

ruyadorno (approved):

PR #986, opened 2025-07-07
gui: focused view

ejkorol:
### Overview

This PR introduces a 'focused' state for the selected item and allows the selected item to take up the entire viewport.

### Changelog
- **updates `sidebar` styles for easier inset styling feat: creates `header` for `dashboard` route feat: updates `explorer-grid` header component feat: creates header for `queries` feat: creates header for `labels` route feat: creates header for `help` routes feat: updates styles on `ui` components feat: integrates new `sidebar` and `header` styles into `layout` feat: improves mobile styles on the header feat: introduces the `linear-menu` component feat: updates components used in `header`**
- **bugfix: check for `importerId` and fix `zIndex` of tooltip**
- **chore: update styles on `explorer-grid`**
- **feat: implements a `focused` view of the selected item**
- **chore: updates props and styles `tabs-overview` and `item-header`**


src/gui/src/components/explorer-grid/selected-item/focused-view/use-focus-state.tsx
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from 'react'
+
+export const useFocusState = (): {
+  focused: boolean
+  setFocused: (focused: boolean) => void
+} => {
+  const [focused, setFocusedState] = useState<boolean>(() => {
+    if (typeof window === 'undefined') return false
```
ruyadorno: what is the goal / intention of this check here?

src/gui/src/state/index.ts
```diff
@@ -71,6 +71,11 @@ const initialState: State = {
   specOptions: undefined,
   stamp: newStamp(),
   theme: localStorage.getItem('vite-ui-theme') as State['theme'],
+  focused: (() => {
+    if (typeof window === 'undefined') return false
```
ruyadorno: same

src/gui/src/components/explorer-grid/selected-item/item-header.tsx
```diff
@@ -91,7 +92,7 @@ const SpecOrigin = ({
             return (
               <DataBadge
                 variant="mono"
-                content={`${item.title}@${item.version}`}
+                content={`${item.title}@${item.version.split('v')[1]}`}
```
ruyadorno: this is likely something we should normalize at the manifest level 🤔

ruyadorno: Implemented here: https://github.com/vltpkg/vltpkg/pull/1015

src/gui/src/components/explorer-grid/selected-item/focused-view/use-focus-state.tsx
darcyclarke: Server-side rendering

src/gui/src/state/index.ts
darcyclarke: See above

src/gui/src/components/explorer-grid/selected-item/focused-view/use-focus-state.tsx
ruyadorno: in this case I think we should go with a more robust feature detection logic, such as: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#feature-detecting_localstorage

also it's worth noting that we are already using local storage in many plces in the app, so we should pb have an abstraction that makes sure it always passes by feature detection first and then it should be used consistently across the app

I think it's fine to land that refactor later but I'd rather keep consistency and if we have not yet used this feature detection pattern we should not make it as a one-off just here

ruyadorno (approved):
LGTM 👍

Issue #968, opened 2025-07-02
[BUG] `vlt config get|set` should accept dot props


ruyadorno:
Similar to `vlt pkg get|set` it should be possible to use dot-props, so that workflows like setting new custom registry origins can then be a one liner, e.g:

```
$ vlt config set registries.local=http://localhost:1337
```

PR #989, opened 2025-07-08
cli-sdk: config cmd handles nested properties

ruyadorno:
This change adds support to using nested properties such as `registries.name` when using the `vlt config` command to get, set and remove properties, e.g:

    $ vlt config set registries.local=http://localhost:1337

    $ vlt config get registries.local
    http://localhost:1337

    $ vlt config del registries.local

Fixes: https://github.com/vltpkg/vltpkg/issues/968


src/cli-sdk/src/commands/config.ts
```diff
@@ -160,6 +161,32 @@ const get = async (conf: LoadedConfig) => {
...
+
+    return dotProp.get(record, subKey) as
+      | string
+      | number
+      | boolean
+      | string[]
+      | undefined
+  }
+
+  // otherwise just get the value directly from the config getter
```
lukekarrys, line 189: It seems like we should use dotProp even if the key does not contain a `.`, or at least in some cases?

For example:

**vlt.json**
```json
{
  "config": {
    "registries": {
      "local": "https://local-registry.local"
    }
  }
}
```

```sh
$ vlt config get registries --config=project
[ 'local=https://local-registry.local' ]
```

I would expect this to return `{ local: 'https://local-registry.local' }` similar to how `vlt pkg get <KEY>` does.

ruyadorno: good call! wasn't too hard to add 👍

lukekarrys (approved):

PR #993, opened 2025-07-08
cli-sdk(pack): replace workspace and catalog specs when packing

lukekarrys:


src/cli-sdk/src/pack-tarball.ts
```diff
@@ -16,181 +18,316 @@ export type PackTarballResult = {
...
+ * @param {Manifest} manifest_ - The manifest to process
+ * @param {LoadedConfig} config - The loaded configuration containing project root, monorepo, and catalog data
+ * @returns {Manifest} The manifest with replaced specs
+ */
+const replaceWorkspaceAndCatalogSpecs = (
+  manifest_: Manifest,
+  config: LoadedConfig,
+): Manifest => {
+  // Create a deep copy of the manifest to avoid modifying the original
+  const manifest = JSON.parse(JSON.stringify(manifest_)) as Manifest
```
ruyadorno: I think structuredClone will be better to use here, wdyt?

```suggestion
  const manifest = structuredClone(manifest_) as Manifest
```

src/cli-sdk/src/pack-tarball.ts
```diff
@@ -16,181 +18,316 @@ export type PackTarballResult = {
...
+    'peerDependencies',
+  ] as const
+
+  for (const depType of depTypes) {
+    const deps = manifest[depType]
+    /* c8 ignore next */
+    if (!deps || typeof deps !== 'object') continue
+
+    const depsObj = deps as Record<string, unknown>
+    for (const [depName, depSpec] of Object.entries(depsObj)) {
```
ruyadorno, line 51: future refactor idea here, let's get the `./dependencies.ts` module out of `@vltpkg/graph` and into its own workspace (so that we don't need to have a dependency on the entire graph library just for reusing this kind of functionality) and reuse the `getDependencies` method https://github.com/vltpkg/vltpkg/blob/5a685ed51dd9d3986104440320ab3625bd1c1c5c/src/graph/src/dependencies.ts#L188-L210

it's already reused in a bunch of places and I think it makes sense to have a standard way of performing this dance

ruyadorno: it needs a little bit more abstraction though as it should take only the manifest instead of relying on the node but anyways, just wanted to flag as a possible directoin in the future, I don't think it's worth investing time on it rn

ruyadorno (approved):

src/cli-sdk/src/pack-tarball.ts
lukekarrys: I was also thinking it would be nice to have a method that read a manifest and would optionally do all the replacements for you. That could live in the same workspace maybe?

ruyadorno: ooohhh maybe 🤔 one thing to also keep in mind is the `normalizeManifest` stuff that recently got beffier with @darcyclarke landing more functions to normalize `contributors` and `maintainers`, so potentially a central place to handle it all

src/cli-sdk/src/pack-tarball.ts
```diff
@@ -16,181 +18,316 @@ export type PackTarballResult = {
...
+ * @param {Manifest} manifest_ - The manifest to process
+ * @param {LoadedConfig} config - The loaded configuration containing project root, monorepo, and catalog data
+ * @returns {Manifest} The manifest with replaced specs
+ */
+const replaceWorkspaceAndCatalogSpecs = (
+  manifest_: Manifest,
+  config: LoadedConfig,
+): Manifest => {
+  // Create a deep copy of the manifest to avoid modifying the original
+  const manifest = structuredClone(manifest_) as Manifest
```
ruyadorno:
```suggestion
  const manifest = structuredClone(manifest_)
```

PR #1015, opened 2025-07-11
types: add manifest.version normalization

ruyadorno:
Adds a normalization step to `manifest.version` values, properly parsing
semver values and asserting they're valid.

### DIFF
--- a/src/types/src/index.ts
+++ b/src/types/src/index.ts
@@ -1,4 +1,5 @@
 import { error } from '@vltpkg/error-cause'
+import { Version } from '@vltpkg/semver'
 
 /** anything that can be encoded in JSON */
 export type JSONField =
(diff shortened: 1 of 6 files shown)


src/types/test/index.ts
```diff
@@ -530,6 +531,158 @@ t.test('normalizeFunding', t => {
...
+    t.equal(result, manifest, 'should return same object reference')
+    t.end()
+  })
+
+  t.test('returns same manifest when version is undefined', t => {
+    const manifest = { name: 'test', version: undefined }
+    const result = normalizeVersion(manifest)
+    t.equal(result, manifest, 'should return same object reference')
+    t.end()
+  })
```
darcyclarke, line 547: @ruyadorno shouldn't we expect to `delete manifest.version` in this case?

src/types/src/index.ts
```diff
@@ -137,6 +138,20 @@ export const normalizeFunding = (
...
 
+/**
+ * Normalize the version field in a manifest.
+ */
+export const normalizeVersion = (
+  manifest: Manifest | ManifestRegistry,
+): Manifest | ManifestRegistry => {
+  if (!manifest.version) {
+    return manifest
+  }
```
darcyclarke, line 149: So you can have `{ "version": undefined }` or `{ "version": null }`? That seems a bit odd. Shouldn't we just delete the value in those cases instead of leaving them intact? Also, what happens with `{ "version": 0 }`? Technically, `0` is a valid semver version but because it's not a string it's going to pass this check & get returned without being normalized to a string (ie. it'll stay a number & future falsey checks will also fail)

src/types/test/index.ts
ruyadorno: it's quite equivalent, I don't think it's necessary or makes any functional difference here 🤔

darcyclarke: It's not the same if you `manifest.toString()` or `Object.entries(manifest)`... the idea is we're "normalizing", ie. "validating", so it would make much more sense to drop keys when they have `undefined`/`null` values which are known not to be valid values for that key.

PR #1043, opened 2025-07-21
server: add config endpoints

ejkorol:
## Overview

This PR introduces new endpoints in the `VltServer` for handling configuration data

#### Changelog
- **feat: moves config logic into new workspace**
- **feat: creates endpoints for `config` in server**
- **tests: writes tests for `config-data` and `handle-request`**
- **feat: passes full LoadedConfig to the server**
- **tests: updates `start-gui` tests**
- **docs: documents config reloading strategy**


src/server/src/config-data.ts
```diff
@@ -0,0 +1,64 @@
...
+    try {
+      await delConfigValue(this.config)
+    } finally {
+      this.config.positionals = originalPositionals
+    }
+  }
+
+  async update(key: string, value: string) {
+    await this.set(key, value)
+  }
```
ruyadorno: let's remove this method and just focus on `set` instead
```suggestion
```

src/server/src/handle-request.ts
ruyadorno, line 1: Let's make sure to run a e2e test to validate that the config reload is reflected after, e.g:
1. Start server
2. validate that a `/config` call is able to retrieve a given config value from the `vlt.json` "config" object
3. use `/config/set` to change a given value and/or add a new one
4. call `/config` pointing to these keys and ensure the value is correct

src/server/src/handle-request.ts
```diff
@@ -150,6 +151,132 @@ export const handleRequest = async (
...
+        return json.ok(res, 'Config value updated successfully')
+      } catch (err) {
+        return json.error(
+          res,
+          'Config update failed',
+          asError(err).message,
+          500,
+        )
+      }
+    }
```
ruyadorno: Let's use only `set`

src/server/test/config-data.ts
```diff
@@ -0,0 +1,263 @@
...
+    1,
+    'positionals restored to original',
+  )
+  t.equal(
+    config.positionals[0],
+    'config',
+    'original positionals restored',
+  )
+  t.end()
+})
```
ruyadorno: update is redundant here too

ruyadorno (changes requested):

src/cli-sdk/src/config/index.ts
ruyadorno, line 1: I like the approach to reload the config here 👍

src/server/src/config-data.ts
```diff
@@ -0,0 +1,147 @@
...
+      const { unload } = (await import(
+        '@vltpkg/vlt-json'
+      )) as VltJsonModule
+      unload('user')
+      unload('project')
+    } catch {
+      /* c8 ignore start */
+      // Continue even if cache clearing fails
+      /* c8 ignore end */
+    }
```
ruyadorno:
```suggestion
    } catch {
      // Continue even if cache clearing fails
    }
```

src/server/src/config-data.ts
```diff
@@ -0,0 +1,147 @@
...
+      }
+
+      return baseConfig
+    }
+
+    // Check if we have fresh data from vlt-json for this key
+    if (configSection && key in configSection) {
+      /* c8 ignore start */
+      return configSection[key]
+      /* c8 ignore end */
```
ruyadorno:
```suggestion
      /* c8 ignore next */
      return configSection[key]
```

src/server/src/config-data.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    if (!key) {
+      // For full config list, merge fresh vlt-json config with the base config
+      const baseConfig = this.list()
+
+      // Overlay fresh vlt-json config values on top of base config
+      if (configSection) {
+        /* c8 ignore start */
+        Object.assign(baseConfig, configSection)
+        /* c8 ignore end */
```
ruyadorno:
```suggestion
        /* c8 ignore next */
        Object.assign(baseConfig, configSection)
```

src/server/src/config-data.ts
```diff
@@ -0,0 +1,147 @@
...
+    try {
+      const { unload } = (await import(
+        '@vltpkg/vlt-json'
+      )) as VltJsonModule
+      unload('user')
+      unload('project')
+    } catch {
+      /* c8 ignore start */
+      // Continue even if cache clearing fails
+      /* c8 ignore end */
```
ruyadorno:
```suggestion
      // Continue even if cache clearing fails
```

src/server/vlt.json
ruyadorno, line 1: is this intentional? I don't think we support internal (as in, internal to workspace folders) `vlt.json` files

vlt.json
```diff
@@ -1,6 +1,5 @@
 {
   "workspaces": ["src/*", "infra/*", "www/*"],
-  "config": {},
```
ruyadorno, line 3: is this intentional?

darcyclarke:
@ruyadorno all great points, I think @ejkorol is going to clean them up & get this over the finish line. 

One outstanding change/comment we should make - @ejkorol - is to remove the comment above `vltJson.unload()` noting that it should only be used for tests (ie. since it was actually one of the critical pieces to making an actual "reload" possible): https://github.com/vltpkg/vltpkg/blob/main/src/vlt-json/src/index.ts#L119

vlt.json
darcyclarke: Can't remember at this point but likely not. I believe I was trying to clean up some changes from running the gui in the project folder & likely removed this unintentionally while trying to remove other arbitrary config values/definitions had been added during testing. That said, why would we want to keep an empty `config` definition?

ruyadorno (approved):

PR #1045, opened 2025-07-24
fix: normalized types

darcyclarke:
- moves work @ejkorol was having to do in the GUI upstream

### DIFF
--- a/src/cli-sdk/tap-snapshots/test/commands/pkg.ts.test.cjs
+++ b/src/cli-sdk/tap-snapshots/test/commands/pkg.ts.test.cjs
@@ -23,7 +23,10 @@ exports[`test/commands/pkg.ts > TAP > init > should init a new package.json file
   "version": "1.0.0",
   "description": "",
   "main": "index.js",
-  "author": "Ruy <ruy@example.com>"
+  "author": {
+    "name": "Ruy",
+    "email": "ruy@example.com"
+  }
 }
 
 `
--- a/src/init/tap-snapshots/test/index.ts.test.cjs
+++ b/src/init/tap-snapshots/test/index.ts.test.cjs
@@ -11,7 +11,10 @@ exports[`test/index.ts > TAP > init > should init a new package.json file 1`] = 
   "version": "1.0.0",
   "description": "",
   "main": "index.js",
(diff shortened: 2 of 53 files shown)


src/types/src/index.ts
```diff
@@ -429,13 +429,36 @@ export type Manifest = {
...
+  Manifest,
+  {
+    bugs: NormalizedBugsEntry[]
+    author: NormalizedContributor
+    contributors: NormalizedContributor[]
+    funding: NormalizedFundingEntry[]
+    keywords: NormalizedKeywords
+  }
+>
+
```
ruyadorno: I think an easier to read and simpler to maintain version of it would be to define `type NormalizedManifest` as being just picking the properties that don't have a "normalized" variation, e.g: `Pick<Manifest, "name", "version", "description", ...>`  and add the normalized values to it, e.g: 
```ts
Pick<Manifest, "..."> & {
  bugs: NormalizedBugsEntry[]
  author: NormalizedContributor
  contributors: NormalizedContributor[]
  funding: NormalizedFundingEntry[]
  keywords: NormalizedKeywords
}
```

src/types/src/index.ts
ruyadorno, line 1: the methods that calls to `normalizeManifest` such as `asManifestRegistry` (currently at L711) needs to stop casting the type to `ManifestRegistry` and return `NormalizedManifest` instead

but actually that leads to another ask, we should probably also have a `NormalizedManifestRegistry` type too, to mimick the `Manifest ~= ManifestRegistry` relationship

ruyadorno (changes requested):

src/cli-sdk/tap-snapshots/test/commands/pkg.ts.test.cjs
ruyadorno, line 1: `vlt pkg` output is now normalized by default

src/init/tap-snapshots/test/index.ts.test.cjs
ruyadorno, line 1: `vlt init` generated `package.json` file is now also normalized by default

ruyadorno:
this is ready to land but I want to see if I can land the benchmark CI first

PR #1069, opened 2025-08-06
graph: reify exit early on empty diff

ruyadorno:
If there are no diffs to be applied by the reify step then it should just exit the earliest as possible.

Refs: https://github.com/vltpkg/statusboard/issues/198

### DIFF
--- a/src/graph/src/diff.ts
+++ b/src/graph/src/diff.ts
@@ -138,4 +138,13 @@ ${lines
   .join('\n')}
 }`
   }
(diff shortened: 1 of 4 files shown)


ruyadorno:
I think we might want to tweak the implementation so that it does write files (`vlt-lock.json`, `node_modules/.vlt-lock.json`) files in case they're missing 🤔

darcyclarke (approved):

PR #1076, opened 2025-08-07
chore: make it easier to run benchmarks locally

lukekarrys:


ruyadorno (approved):
Little info dump from my pairing session with @lukekarrys: The `clean` and `lockfile` variations are made slower here since there was a problem with the previous script trying to remove the cache folder in between runs, that was causing these variation to "speed up" since they were reaching for a warm cache when it should not be there to begin with, that's why we see a degradation of performance numbers here 👍 LGTM

PR #1080, opened 2025-08-07
chore(benchmark): set binary through hyperfine parameter list

lukekarrys:
This has no effect on CI benchmarks but now the `vlt` binary is supplied to hyperfine via `--parameter-list`. This allows for locally running the benchmarks against a baseline to help debug performance.

Here's how I'm using it locally with two local copies of the `vltpkg/vltpkg` repo:

```sh
export BENCH_BINARY="$(pwd)/scripts/bins/bundle/vlt,/Users/lukekarrys/projects/vltpkg/vltpkg-clean/scripts/bins/bundle/vlt"
bash ./infra/cli-benchmarks/scripts/benchmark.sh abbrev clean
```


ruyadorno:
shouldn't the `BENCH_BINARY` variable be used in the `./scripts/variations/common.sh` `BENCH_COMMAND_VLT` value?

I'm not really following how this is supposed to work 😅

ruyadorno (approved):

PR #1085, opened 2025-08-10
gui: query builder

ejkorol:
https://github.com/user-attachments/assets/92c251a6-7885-448f-b3c4-bdabb9415ff4

This PR introduces a new feature into the gui, allowing users to visually create queries using a new ‘query builder’ interface.

The query builder state is ‘open’ by default, when the input is initially focused. Optionally allowing a user to toggle whether the query builder should be visible on consecutive focuses.

A combobox present serves as the search / selection interface for building the queries featuring human readable vocabulary for the selectors you want. Descriptions of each selector is displayed on hover under the combobox.

### Selector Constants
Selector constants in `@/lib/constants/selectors` (previously used for the `/help/selectors` route) are updated to contain argument, combinator and attribute information. This will also be used for the `query-builder` to effectively treat the arguments / operators in here as the source of truth for building queries.

Additionally we kill the `socket-selectors` previously used for the security selector information, and prefer to colocate those pseudo selectors in the same file as the rest of the selectors.

### Query Bar
The `query-bar` is updated to export it’s ID property (used for creating a portal from the `query-builder` to the `query-bar`) and accept new props which include it’s wrapper `className` and new `onFocus` and `onBlur` props.

Additionally, because of the query builder, we manually define a ‘focus’ ring for the input element using `borders` to signify it’s state.

### Query Builder
A new directory containing the query builder components / interfaces is created in `@/components/query-builder/`. The `index` exports the main `query-builder` ui as well as a `button` and conditional which renders the `query-builder`.

The `query-builder` is attached to the `query-bar` via a portal, and renders a ‘focus guard’ behind  the `query-builder` and `query-bar` which handles interactions when the user clicks outside the components.

The `query-builder` utilizes `ast` helpers which serve as wrappers around the `parse` method from `@vltpkg/dss-parser` to build the queries.

We make round trip conversions `input -> parse -> nodes -> serialize -> input` whenever the `input` (query) changes.

#### `ast-interface.ts` and `ast-tokenize.ts`
To effectively build queries, we utilize the `parse` method from `dss-parser` to turn the `query` state into a _postcss_ `ast`. Then we ‘transform’ this `ast` into a ui-friendly representation which serves as a helpful wrapper / abstraction over the otherwise awkward css `ast` to iterate and manipulate the nodes / tokens simply.

We utilize this ‘ui-friendly’ ast to then manipulate the nodes and build queries as necessary.

#### `ast-stringify.ts`
We make a round trip conversion back into a query string by creating a new _postcss_ `ast` and appending the ui-nodes back into `postcss nodes` then calling the `toString` method to turn this back into a query string.

### State
The global store is updated to contain the related `query-builder` states.


src/gui/src/components/query-builder/ast-tokenize.ts
```diff
@@ -0,0 +1,212 @@
...
+    !/[>,~]$/.test(trimmed) &&
+    (result[result.length - 1]?.type !== 'combinator' ||
+      (result[result.length - 1] as UiNode & { value?: string })
+        .value !== ' ')
+  ) {
+    result = [...result, { type: 'combinator', value: ' ' }]
+  }
+
+  return result
+}
```
ruyadorno: Let's ideally remove the regex checks here, also some checks seem to be duplicated, ideally we're relying only on the ast objects, we don't need to be validating and / or checking the query string at the same time

src/gui/src/components/query-builder/ast-tokenize.ts
```diff
@@ -0,0 +1,212 @@
...
+// Preserve trailing draft tokens while typing (comma and descendant space)
+const appendDraftSuffixTokens = (
+  query: string,
+  tokens: UiNode[] | undefined,
+): UiNode[] | undefined => {
+  if (!tokens) return tokens
+  let result = tokens
+
+  if (
+    /,\s*$/.test(query) &&
```
ruyadorno: this can maybe be?

```suggestion
    query.indexOf(',') > -1 &&
```

src/gui/src/components/query-builder/ast-tokenize.ts
```diff
@@ -0,0 +1,212 @@
...
+    result[result.length - 1]?.type !== 'comma'
+  ) {
+    result = [...result, { type: 'comma', value: ',' }]
+  }
+
+  const trimmed = query.replace(/\s+$/, '')
```
ruyadorno: maybe?

```suggestion
  const trimmed = query.trim()
```

src/gui/src/components/query-builder/ast-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+export const createSelector = () => selector({ value: '' })
+export const createRoot = () => root({ value: '' })
+
+export type PostCSSLeaf =
+  | ReturnType<typeof tag>
+  | ReturnType<typeof id>
+  | ReturnType<typeof attribute>
+  | ReturnType<typeof combinator>
+  | ReturnType<typeof pseudo>
+  | ReturnType<typeof string>
```
ruyadorno: it would be nice to have this live in the `@vltpkg/dss-parser` library so that other parts of the code could potentially reuse helpers like this

src/gui/src/components/query-builder/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+        <Tooltip>
+          <TooltipTrigger asChild>
+            <Button
+              onClick={onClick}
+              className="duration-250 flex h-[1.5rem] w-[1.5rem] items-center justify-center rounded-sm bg-input p-0 text-muted-foreground transition-colors hover:bg-neutral-300 hover:text-accent-foreground dark:hover:bg-neutral-700 [&_svg]:pointer-events-none [&_svg]:size-4 [&_svg]:shrink-0">
+              <UnfoldVertical />
+            </Button>
+          </TooltipTrigger>
+          <TooltipPortal>
+            <TooltipContent>View query builder</TooltipContent>
```
ruyadorno: should toggle the text content here for when the query builder is open

ruyadorno (approved):

PR #1087, opened 2025-08-11
graph: add lockfile only option

ruyadorno:
Add a `--lockfile-only` option to the CLI and the support to a `lockfileOnly` option in the `@vltpkg/graph` library so that installs are limited to only saving to the `vlt-lock.json` and `package.json` files while avoiding any `node_modules` related file system operations.

Refs: https://docs.npmjs.com/cli/v11/commands/npm-install#package-lock-only
Refs: https://pnpm.io/cli/install#--lockfile-only


ruyadorno:
Will also need to write to importers `package.json` files.

darcyclarke (approved):

Issue #748, opened 2025-05-06
[CLEANUP] Refactor virtual Graph loading API


ruyadorno:
We need to bring the `src/gui/src/state/load-graph.ts` module into `@vltpkg/graph/browser` exposed as a `loadVirtualGraph` method. Along with that refactor we may as well define more concrete isomorphic class implementations of the `GraphLike`, `EdgeLike` and `NodeLike` types.

That will reduce the amount of code / boilerplate required on the GUI to set up the loaded Graph but also make it so that this isomorphic structure is reusable in other runtimes (e.g: Cloudflare workers)

darcyclarke:
@ruyadorno is this done/can this be closed out for a different issue? We have peer-dep related issues.

PR #1094, opened 2025-08-13
graph: fix ideal graph determinism

ruyadorno:
Various fixes to the graph library and some of its auxiliary libraries with the goal of insuring installs are deterministic.

- Tweaks src/graph/src/ideal/append-nodes.ts so that fetching dependency manifests are deterministic independent of concurrency while fetching remote manifests.
- Added an inheritedRegistry setter to the `Spec` object.
- Added references to the inherited registry to nodes being loaded from the actual file.
- Fixed `@vltpkg/dep-id` hydration, adding missing spec options forwarding and always referencing a name.

Refs: https://github.com/vltpkg/vltpkg/issues/748


src/dep-id/src/browser.ts
ruyadorno, line 1: this is a bugfix, the `hydrate` method was simply not forwarding the spec options and as such would be unable to properly handle custom registries, etc

src/graph/src/actual/load.ts
ruyadorno, line 1: I'm introducing the concept of an `inheritedRegistry` in the `Spec` object as a setter so that it can properly handle setting that value to be used when a spec would be normally just using to the default registry value

src/graph/src/graph.ts
```diff
@@ -40,8 +42,18 @@ const getResolutionCacheKey = (
...
+  // and in the case of a registry, what registry it is from.
+  const typePrecisionKey =
+    f.registry ?
+      `${cacheKeySeparator}registry` +
+      `${cacheKeySeparator}${f.registry}`
+    : f.gitRemote ?
+      `${cacheKeySeparator}git` + `${cacheKeySeparator}${f.gitRemote}`
+    : `${cacheKeySeparator}${f.type}`
+  const modifierSuffix = `${cacheKeySeparator}${queryModifier}`
+  return fromPrefix + String(f) + typePrecisionKey + modifierSuffix
```
ruyadorno, line 56: these cache keys needed to be aware of more information such as registries so that it won't just retrieve a package with the same name that came from a different registry

src/graph/src/graph.ts
```diff
@@ -393,6 +394,18 @@ export class Graph implements GraphLike {
...
+    const f = getResolutionCacheKey(
+      spec.final,
+      fromNode.location,
+      extra || '',
+    )
+    this.resolutions.set(f, toNode)
+    const rrev = this.resolutionsReverse.get(toNode) ?? new Set()
+    rrev.add(f)
+    this.resolutionsReverse.set(toNode, rrev)
+
```
ruyadorno, line 408: moved from `addNode` to `placePackage` which simplified a bit the logic but most importantly allows for using the proper `getResolutionCacheKey` method instead of relying on the stringified spec as a key.

src/graph/src/graph.ts
```diff
@@ -463,6 +476,50 @@ export class Graph implements GraphLike {
...
+        edge.spec.final,
+        edge.from.location,
+        queryModifier,
+      )
+
+      this.resolutions.set(resolutionKey, node)
+      this.resolutionsReverse.get(node)?.add(resolutionKey)
+    }
+  }
+
```
ruyadorno, line 522: added this method so that all resolution caches can be populated when loading from a lockfile, this way graph resolutions that require a full build-ideal step should benefit form that cache lookup to reuse packages already present in the lockfile

src/graph/src/ideal/append-nodes.ts
ruyadorno, line 1: This is the main change, there's a lot of line changes which are indentation only so either activating the "Hide whitespaces" option in the GitHub review or using `git diff -w` will help reduce the number of diff lines to read.
The main change is letting only the package manifest fetch happen in parallel but then wait for all of their promises to resolve and sort them alphabetically.

src/graph/src/lockfile/load-edges.ts
```diff
@@ -48,13 +48,18 @@ export const loadEdges = (
         },
       })
     }
+
+    // sets a registry for this spec to inherit from
+    spec.inheritedRegistry = from.registry
+
```
ruyadorno, line 54: implement in the lockfile logic the same logic present in the actual load module

ruyadorno:
Closing due to concerns with the performance impact: https://github.com/vltpkg/vltpkg/pull/1094#issuecomment-3182028354

ruyadorno:
Follow up that fixes the minor issues that were also addressed in this PR: https://github.com/vltpkg/vltpkg/pull/1107

PR #1105, opened 2025-08-15
fix(cli-sdk): use ink's waitUntilExit instead of forcing unmount

lukekarrys:
Also update to ink@6 and react@19. Together this fixes some rendering bugs
in the install output due to some race conditions when install steps happen
syncronhously, e.g. when install bails early due to no diff.


lukekarrys:
It seems like updating `cli-sdk` to use `react@19` is affecting linting in the `gui` workspace. Maybe because that is the only other workspace that uses `react`? Some peer dependency issue probably.

darcyclarke (approved):

PR #1126, opened 2025-08-22
fix: create path for user level config if it doesnt exist

darcyclarke:
`vlt config set ... --config=user` silently fails today if the `vlt.json` doesn't exists as it's not able to create it in the path (because the `vlt` dir also doesn't exist in the path to the user-level config).

### DIFF
--- a/src/vlt-json/test/index.ts
+++ b/src/vlt-json/test/index.ts
@@ -339,8 +339,14 @@ t.test('do not walk up past a .git folder', async t => {
 })
 
 t.test(
-  'create directories when saving to non-existent path',
+  'create directories when saving user config to non-existent path',
   async t => {
+    class MockXDG extends XDG {
+      config() {
+        return resolve(dir, 'a/d/vlt.json')
+      }
+    }
+
     const dir = t.testdir({
       a: {
         b: {
@@ -354,18 +360,19 @@ t.test(
     const { load, save, find } = await t.mockImport<
       typeof import('../src/index.ts')
     >('../src/index.ts', {
+      '@vltpkg/xdg': { XDG: MockXDG },
       'node:os': t.createMock(await import('node:os'), {
-        homedir: () => t.testdirName,
+        homedir: () => resolve(dir, 'a/d'),
       }),
     })
 
-    // The find function should return a path that doesn't exist yet
-    const configPath = find()
-    t.equal(configPath, resolve(dir, 'a/b/vlt.json'))
+    // The find function should return a path to a vlt.json file that doesn't exist yet
+    const configPath = find('user')
+    t.equal(configPath, resolve(dir, 'a/d/vlt.json'))
 
     // Directory doesn't exist initially
     t.equal(
-      await lstat(resolve(dir, 'a/b/vlt.json')).catch(() => false),
+      await lstat(resolve(dir, 'a/d')).catch(() => false),
       false,
     )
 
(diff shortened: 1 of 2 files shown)


src/vlt-json/test/index.ts
```diff
@@ -337,3 +337,53 @@ t.test('do not walk up past a .git folder', async t => {
...
+        },
+      },
+    })
+    t.chdir(resolve(dir, 'a/b/c'))
+
+    const { load, save, find } = await t.mockImport<
+      typeof import('../src/index.ts')
+    >('../src/index.ts', {
+      'node:os': t.createMock(await import('node:os'), {
+        homedir: () => t.testdirName,
```
ruyadorno: tap's `t.testdirName` will pb not be the same as the returned `dir` from calling `t.testdir`, let's not mix them as that can easily get confusing

```suggestion
        homedir: () => dir,
```

src/vlt-json/test/index.ts
```diff
@@ -337,3 +337,53 @@ t.test('do not walk up past a .git folder', async t => {
...
+
+    // The find function should return a path that doesn't exist yet
+    const configPath = find()
+    t.equal(configPath, resolve(dir, 'a/b/vlt.json'))
+
+    // Directory doesn't exist initially
+    t.equal(
+      await lstat(resolve(dir, 'a/b/vlt.json')).catch(() => false),
+      false,
+    )
```
ruyadorno: It says it's checking for the directory but it's actually pointing to the `vlt.json` file in this `lstat` call

```suggestion
    // Directory doesn't exist initially
    t.equal(
      await lstat(resolve(dir, 'a/b')).catch(() => false),
      false,
    )
```

ruyadorno (dismissed):

src/vlt-json/test/index.ts
ruyadorno: actually we should make it point to a folder `a/d/` that will not exist then:

```suggestion
        homedir: () => resolve(dir, 'a/d'),
```

src/vlt-json/test/index.ts
```diff
@@ -337,3 +337,53 @@ t.test('do not walk up past a .git folder', async t => {
...
+
+    const { load, save, find } = await t.mockImport<
+      typeof import('../src/index.ts')
+    >('../src/index.ts', {
+      'node:os': t.createMock(await import('node:os'), {
+        homedir: () => t.testdirName,
+      }),
+    })
+
+    // The find function should return a path that doesn't exist yet
```
ruyadorno:
```suggestion
    // The find function should return a path to a vlt.json file that doesn't exist yet
```

src/vlt-json/test/index.ts
```diff
@@ -337,3 +337,53 @@ t.test('do not walk up past a .git folder', async t => {
...
+
+    const assertBool = (b: unknown): asserts b is boolean => {
+      if (b !== !!b) throw new Error('expected boolean')
+    }
+
+    // Load should return undefined since file doesn't exist
+    t.equal(load('testField', assertBool), undefined)
+
+    // Save should create the directory and file
+    save('testField', true)
```
ruyadorno: I think we'll need to also set the which value to point to user:
```suggestion
    // Save should create the directory and file
    save('testField', true, 'user')
```
but then we might need to tweak something else in order to have the "user" point to the `a/d/` mock folder we want to use

src/vlt-json/test/index.ts
```diff
@@ -337,3 +337,53 @@ t.test('do not walk up past a .git folder', async t => {
...
+    const { load, save, find } = await t.mockImport<
+      typeof import('../src/index.ts')
+    >('../src/index.ts', {
+      'node:os': t.createMock(await import('node:os'), {
+        homedir: () => t.testdirName,
+      }),
+    })
+
+    // The find function should return a path that doesn't exist yet
+    const configPath = find()
```
ruyadorno: Find should likely point to the "user" profile

```suggestion
    const configPath = find('user')
```

Issue #1088, opened 2025-08-11
[CHORE]  Cleanup Lockfile


darcyclarke:
### 📋 Description

We're reaching out to registry metadata even with a lockfile in place which is incorrect based on our planned package manager architecture.

### 📂 Category

Performance Optimization - Improve speed/efficiency

### 🎯 Motivation

- we look super slow compared to other package managers in the lockfile-related scenarios, ref: https://vltpkg.github.io/benchmarks/

### 🎌 Scope of Work

- graph resolution (install/ci/update)
- lockfile

### 📊 Impact Level

Critical - Major changes, needs careful planning

### ⏰ Urgency

Critical - Needs immediate attention

PR #1139, opened 2025-08-29
tar: add piscina for worker thread pool

ruyadorno:
Manages the worker thread pool using `piscina` instead of the custom pool implementation.

Refs: https://github.com/vltpkg/vltpkg/issues/1088


ruyadorno:
I'll close this one for now as there are no plans to iterate on it in the foreseeable future nor there's much chance it can deliver performance improvements.

PR #1147, opened 2025-08-30
graph: add lockfile platform info

ruyadorno:
Add the reify-required platform info so that reify does not have an implicity dependency on the manifest data.


darcyclarke (approved):

ruyadorno:
in its current state optional deps are back at being installed for the wrong platform, I'm investigating what the problem might be

ruyadorno:
fixed

PR #1176, opened 2025-09-10
fix(server): port forwarding/dashboard loading in codespace envs

darcyclarke:
- modified the host/port config to accommodate codespace envs where the logged url is wrong/stuck on `localhost` & `dashboard.json` fails to load
- added graceful fallback for `homedir()` if it throws (rare but can happen)

### DIFF
--- a/src/server/src/dashboard.ts
+++ b/src/server/src/dashboard.ts
@@ -1,6 +1,6 @@
 import { writeFileSync } from 'node:fs'
 import { homedir } from 'node:os'
-import { resolve } from 'node:path'
+import { resolve, dirname } from 'node:path'
 import { getUser } from '@vltpkg/git'
 import { getAuthorFromGitUser } from '@vltpkg/init'
 import { getReadablePath } from './get-readable-path.ts'
@@ -69,7 +69,7 @@ export class Dashboard {
       } catch {
         // In restricted environments (like locked-down Codespaces),
         // homedir() might fail. Fall back to current working directory.
-        this.dashboardRoot = [process.cwd()]
+        this.dashboardRoot = [dirname(process.cwd())]
       }
     }
     this.scurry = scurry
--- a/src/server/src/dashboard.ts
+++ b/src/server/src/dashboard.ts
@@ -63,7 +63,15 @@ export class Dashboard {
     } = options
     this.packageJson = packageJson
     this.dashboardRoot = dashboardRoot
-    if (!this.dashboardRoot.length) this.dashboardRoot = [homedir()]
+    if (!this.dashboardRoot.length) {
+      try {
+        this.dashboardRoot = [homedir()]
+      } catch {
+        // In restricted environments (like locked-down Codespaces),
+        // homedir() might fail. Fall back to current working directory.
+        this.dashboardRoot = [process.cwd()]
+      }
+    }
     this.scurry = scurry
     this.publicDir = publicDir
   }
--- a/src/server/src/get-readable-path.ts
+++ b/src/server/src/get-readable-path.ts
@@ -1,4 +1,5 @@
 import { homedir } from 'node:os'
+import { dirname } from 'node:path'
 
 let home: string
 try {
@@ -6,7 +7,7 @@ try {
 } catch {
   // In restricted environments (like locked-down Codespaces),
   // homedir() might fail. Fall back to current working directory.
-  home = process.cwd()
+  home = dirname(process.cwd())
 }
 
 export const getReadablePath = (path: string) =>
--- a/src/server/src/get-readable-path.ts
+++ b/src/server/src/get-readable-path.ts
@@ -1,6 +1,13 @@
 import { homedir } from 'node:os'
 
-const home = homedir()
+let home: string
+try {
+  home = homedir()
+} catch {
+  // In restricted environments (like locked-down Codespaces),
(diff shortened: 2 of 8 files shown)


src/server/src/dashboard.ts
```diff
@@ -63,7 +63,15 @@ export class Dashboard {
...
-    if (!this.dashboardRoot.length) this.dashboardRoot = [homedir()]
+    if (!this.dashboardRoot.length) {
+      try {
+        this.dashboardRoot = [homedir()]
+      } catch {
+        // In restricted environments (like locked-down Codespaces),
+        // homedir() might fail. Fall back to current working directory.
+        this.dashboardRoot = [process.cwd()]
+      }
+    }
```
ruyadorno, line 74: this breaks a little the main logic for the dashboard root, I'd rather default to an empty array

src/server/src/get-readable-path.ts
```diff
@@ -1,6 +1,13 @@
...
 
-const home = homedir()
+let home: string
+try {
+  home = homedir()
+} catch {
+  // In restricted environments (like locked-down Codespaces),
+  // homedir() might fail. Fall back to current working directory.
+  home = process.cwd()
+}
```
ruyadorno, line 11: this is not necessary in the case we default to an empty array as suggested above

src/server/src/index.ts
```diff
@@ -118,7 +118,18 @@ export class VltServer extends EventEmitter<{
...
+    /* c8 ignore start */
+    if (
+      process.env.CODESPACE_NAME &&
+      process.env.GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN
+    ) {
+      this.#rootAddress = `https://${process.env.CODESPACE_NAME}-${this.port}.${process.env.GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN}/`
+    } else {
+      this.#rootAddress = `http://localhost:${this.port}/`
+    }
+    /* c8 ignore stop */
```
ruyadorno, line 132: this is great! 👍 I do wonder about supporting more environments though, maybe there's the potential to have a small package to aggregate a bunch of those services in the future 🤔

src/server/src/read-project-folders.ts
```diff
@@ -4,7 +4,14 @@ import { callLimit } from 'promise-call-limit'
...
 const limit = Math.max(availableParallelism() - 1, 1) * 8
-const home = homedir()
+let home: string
+try {
+  home = homedir()
+} catch {
+  // In restricted environments, homedir() might fail.
+  // Fall back to current working directory.
+  home = process.cwd()
+}
```
ruyadorno, line 15: same as mentioned before, I think the better solution is to have no paths in this case, I'd rather change the logic to return early an empty array if we see that `userDefinedProjectPaths.length === 0`

src/server/src/read-project-folders.ts
```diff
@@ -105,9 +118,13 @@ const collectResult = async (
...
-  const statPackageJson = await scurry.lstat(
-    `${resolved}/package.json`,
-  )
+  let statPackageJson
+  try {
+    statPackageJson = await scurry.lstat(`${resolved}/package.json`)
+  } catch {
+    // Ignore files that can't be accessed (permission denied, etc.)
+    statPackageJson = null
+  }
```
ruyadorno: I find this type of situation to be the perfect usage of the shorter form:

```suggestion
const statPackageJson = await scurry.lstat(
  `${resolved}/package.json`,
).catch(() => {})
```

ruyadorno (changes requested):

ruyadorno:
Defaulting to `process.cwd()` is going to conceptually break the logic for `:host-context(localhost)` that I'm currently working on, if the goal is to be more resilient in these special environments I think a better solution is to skip a bunch of this logic.

src/server/src/dashboard.ts
vltsandbox: @ruyadorno you'd rather have nothing vs. something? Like, the dashboard then wouldn't have the project you initialized in even(?)

darcyclarke: @ruyadorno you'd rather have nothing vs. something? Like, the dashboard then wouldn't have the project you initialized in even(?)

src/server/src/index.ts
darcyclarke: 100% agree & was going to make a note about this/more envs but just wanted to unblock myself (ie. let's punt this to another issue) - this was very annoying to not be able to run the gui in Codespaces for testing malware/sandboxing

src/server/src/dashboard.ts
ruyadorno: right, does this kind of environment have a folder structure available? if we have at least a parent folder it would be better to use that instead then, e.g:

assuming the project gets cloned / mounted on a `/project-name` folder we could make it use `/` then

darcyclarke: right, so the change there would be `dirname(process.cwd())`(?)

ruyadorno: SGTM 👍

ruyadorno: and do the same in the other 2 places that is handling this fallback below

ruyadorno (approved):

PR #1232, opened 2025-10-01
package-info: add manifest resolution cache

ruyadorno:
Add an extra layer of cache in front of the manifest retrieval function in `@vltpkg/package-info`, this speeds up cached installs by avoiding loading the very large packument files when a given resolution to a manifest request is already known.


ruyadorno:
note: this improves significantly the `cache` benchmarks variation which unfortunately is currently not present in the CI benchmarks

lukekarrys (approved):

PR #1236, opened 2025-10-06
package-info: fix single manifest request

ruyadorno:
Strips possible leading characters from returned `registrySpec` values in the single manifest request method.


ruyadorno:
unfortunately we're having a bootstrap issue here since `npx vlt` is being affected by the very bug this PR is fixing 😞 

I think the simpler solution here is going to be to just land despite the failed targets

Issue #1148, opened 2025-08-30
[BUG] Add `bins` info to lockfile


ruyadorno:
Similar to https://github.com/vltpkg/vltpkg/pull/1147 we should store bin info to the lockfile so that we can run a full reify with no manifests.

PR #1250, opened 2025-10-10
graph: add bin info to lockfile

ruyadorno:
When reifying from a lockfile we need to store the bin info in order to properly symlink and chmod the bin files at the once all files have been extracted.

Fixes: https://github.com/vltpkg/vltpkg/issues/1148


ruyadorno:
@lukekarrys heads up, here I'm replacing the "Bootstrap" phase of the `vlt.yml` with a `pnpm install` for now, which is def safer to use, we should likely pin it to use our `vlt@1.0.0` only after it's out of beta.

PR #1256, opened 2025-10-12
gui: package search

ejkorol:


src/gui/src/lib/package-search.ts
```diff
@@ -0,0 +1,155 @@
...
+  const queryParams = new URLSearchParams(
+    params as Record<string, string>,
+  ).toString()
+  const url = `${endpoint}?${queryParams}`
+
+  const res = await fetch(url, {
+    method: 'GET',
+    headers: {
+      'User-Agent': 'vlt-gui-search/1.0',
+    },
```
ruyadorno: Setting an user agent header triggers CORS errors in Firefox

src/gui/src/state/types.ts
```diff
@@ -35,6 +35,12 @@ export type Action = {
     queryInputFocused: State['queryInputFocused'],
   ) => void
   updateFocused: (focused: State['focused']) => void
+  updateIsExternalPackage: (
+    isExternalPackage: State['isExternalPackage'],
+  ) => void
+  updateExternalPackageSpec: (
+    externalPackageSpec: State['externalPackageSpec'],
+  ) => void
```
ruyadorno, line 43: I think it makes a lot more sense to save an object with `name` and `version` here

src/gui/src/utils/extract-repo-org.ts
```diff
@@ -0,0 +1,16 @@
+export const extractRepoOrg = (
+  repoUrl: string,
+): string | undefined => {
+  try {
+    const url = new URL(repoUrl)
+    if (url.hostname === 'github.com') {
```
ruyadorno, line 6: 🤔 could it also potentially be `www.github.com` ?

PR #1264, opened 2025-10-20
fix: issue/1194

### LINKED ISSUE #1194: [BUG] Commands don't show flags/options for `--help`
- super annoying/sort of insane to not have output

Image

eryue0220:
Fix #1194 

Here is the output

bd938241-fc48-4ac7-9c7a-2e0f15202318
206a678e-3662-403d-a3c2-ef32fd259ac1
dd98f1ee-e4f6-4477-8673-e560691f3ac7


ruyadorno (changes requested):
Thank you so much for taking the time to send this contribution @eryue0220!

Unfortunately in its current state the changes proposed here are not fixing the exact problem described in https://github.com/vltpkg/vltpkg/issues/1194, which is to provide the exact supported options for each command.

The correct way to have those showing up in the help is to properly populate the `options` property send in the `commandUsage` call for the `usage` exported property of each command module. Good command examples at the moment would be https://github.com/vltpkg/vltpkg/blob/main/src/cli-sdk/src/commands/build.ts and https://github.com/vltpkg/vltpkg/blob/main/src/cli-sdk/src/commands/query.ts, you can observe that these commands correctly display supported options when you run help for them, e.g:

```
$ vlt build --help
Usage:
  vlt build [query]
  vlt build [--target=<query>]

Build the project based on the current dependency graph.

This command processes the installed packages in node_modules and runs any
necessary build steps, such as lifecycle scripts and binary linking.
The build process is idempotent and will only perform work that is actually
needed based on the current state of the dependency graph.
Use --target option or provide a query as a positional argument to filter
packages using DSS query language syntax, otherwise it will target all packages
with scripts (:scripts) by default.

  Aliases

    ​b

  Options

    target
      Query selector to filter packages using DSS syntax.

      ​--target=<query>
```

`vlt help query` is an even better example since it has multiple options document but for the sake of brevity I prefer to use `vlt build` above for the example.

It might be a bit too much to ask for an external contributor to work on it but if you're motivated, a few of the commands have more options documented in their docs page (I know it for a fact since I recently expanded the docs for https://docs.vlt.sh/cli/commands/install) so that could be a good starting point. Otherwise it's more about chasing what options from https://github.com/vltpkg/vltpkg/blob/main/src/cli-sdk/src/config/definition.ts are used where and properly documenting these on each command using the `commandUsage` API.

Personally I would recommend closing this PR and starting a new one with that new direction in mind but of course you're free to also just force push this same branch to continue the work from a clean slate.

All the best and thanks again for the contribution! ⚡

eryue0220:
> Thank you so much for taking the time to send this contribution @eryue0220!
> 
> Unfortunately in its current state the changes proposed here are not fixing the exact problem described in #1194, which is to provide the exact supported options for each command.
> 
> The correct way to have those showing up in the help is to properly populate the `options` property send in the `commandUsage` call for the `usage` exported property of each command module. Good command examples at the moment would be https://github.com/vltpkg/vltpkg/blob/main/src/cli-sdk/src/commands/build.ts and https://github.com/vltpkg/vltpkg/blob/main/src/cli-sdk/src/commands/query.ts, you can observe that these commands correctly display supported options when you run help for them, e.g:
> 
> ```
> $ vlt build --help
> Usage:
>   vlt build [query]
>   vlt build [--target=<query>]
> 
> Build the project based on the current dependency graph.
> 
> This command processes the installed packages in node_modules and runs any
> necessary build steps, such as lifecycle scripts and binary linking.
> The build process is idempotent and will only perform work that is actually
> needed based on the current state of the dependency graph.
> Use --target option or provide a query as a positional argument to filter
> packages using DSS query language syntax, otherwise it will target all packages
> with scripts (:scripts) by default.
> 
>   Aliases
> 
>     ​b
> 
>   Options
> 
>     target
>       Query selector to filter packages using DSS syntax.
> 
>       ​--target=<query>
> ```
> 
> `vlt help query` is an even better example since it has multiple options document but for the sake of brevity I prefer to use `vlt build` above for the example.
> 
> It might be a bit too much to ask for an external contributor to work on it but if you're motivated, a few of the commands have more options documented in their docs page (I know it for a fact since I recently expanded the docs for https://docs.vlt.sh/cli/commands/install) so that could be a good starting point. Otherwise it's more about chasing what options from https://github.com/vltpkg/vltpkg/blob/main/src/cli-sdk/src/config/definition.ts are used where and properly documenting these on each command using the `commandUsage` API.
> 
> Personally I would recommend closing this PR and starting a new one with that new direction in mind but of course you're free to also just force push this same branch to continue the work from a clean slate.
> 
> All the best and thanks again for the contribution! ⚡

Thanks for your explain, I'll continue to try on this.

ruyadorno:
I'm going to close this PR as we're cleaning up stale PRs / issues, feel free to open a new one in the future 😊

PR #1275, opened 2025-11-04
semver: avoid usage of instanceof

ruyadorno:
Tests were breaking in a completely unrelated usage of the `parse()` method, changing to a more reliable check.


ruyadorno:
Note: I locally ran the `benchmark` script for `src/semver` to validate there are no performance regressions associated with the change.

Issue #1294, opened 2025-11-19
[BUG] json output is including extraneous results


ruyadorno:
Using json output when filtering packages using the `:scripts` selector yields extraneous results such as root package or workspaces.

PR #1296, opened 2025-11-19
graph: fix json output

ruyadorno:
JSON output should only list edges and nodes that are actually selected as part of a query.

Fixes: https://github.com/vltpkg/vltpkg/issues/1294


ruyadorno:
@cursoragent run `pnpm snap` in the `src/cli-sdk` folder and commit the result

PR #1304, opened 2025-11-26
graph: remove extra labels from mermaid output

ruyadorno:
Further minifies the output of the mermaid renderer by removing repeated instances of node names and extra lines.


ruyadorno:
@darcyclarke no perf concerns here, given that the benchmarks still reach out to the network it's unfortunately still possible to have flaky/random results due to fetch calls, etc

after refreshing the PR the benchmark report looks good now 👍

PR #1306, opened 2025-11-26
graph: internal hoist optimization

ruyadorno:
Added a few optimizations to the internal hoist logic that were surfaced during a debugging session trying to speed up the cache+lockfile install type.


ruyadorno:
This is not yielding the potential performance improvements expected

PR #1332, opened 2025-12-08
dep-id: normalize joinDepIDTuple default registry

ruyadorno:
When using the `joinDepIDTuple` method of `@vltpkg/dep-id` using the default registry as a custom registry value - that's possible when declaring dependencies such as:

```
  "foo": "registry:https://registry.npmjs.org#foo@1.0.0"
```

The final DepID value should be normalized to use the default registry name instead.

Refs: https://github.com/vltpkg/vltpkg/pull/1330

### DIFF
--- a/src/dep-id/src/browser.ts
+++ b/src/dep-id/src/browser.ts
@@ -1,6 +1,7 @@
 import { error } from '@vltpkg/error-cause'
 import {
   currentDefaultRegistryName,
+  defaultRegistry,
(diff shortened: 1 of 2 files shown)


ruyadorno:
Benchmark results are not looking good and this is very much an edge case, I think I'll back away from it at this point in time but we may want to revisit this in the future if we have any problems related to this small discrepancy.

PR #1333, opened 2025-12-08
[bugfixes] gui: improve performance in `selected-item` and `external-info`

ejkorol:
# Overview

This PR addresses a lot of performance improvements in the GUI in relation to the `selected-item` component and the `external-info` module.

## Improvements

Before | After | Improvement
-- | -- | --
~20+ re-renders per selected item | 2 re-renders per selected item | ~90% reduction
Duplicate packument fetches, duplicate publisher logic | Each API called exactly once | Eliminated duplicates
Streaming updates could show stale data if user changed selection mid-stream | Single atomic update eliminates race conditions
UI flickered as partial data streamed in | Clean single update with complete data
Multiple layout shifts as content populated | One layout adjustment when data arrives

### Before
https://github.com/user-attachments/assets/ff327844-c5a8-4977-9d79-932d9be16193

### After
https://github.com/user-attachments/assets/2eb40119-5437-4d8b-bc2a-185122d93687

## What were the issues?

In our original implementation of the GUI, we prioritized getting data back to the UI as quickly as possible, so the UI can at least present something while the rest of the data was being loaded / fetched. We did this by streaming UI updates directly into the state in the `selected-item/context` tree. With all the race conditions and state updates, this produces layout thrashing and popcorning in the UI as data was loaded in incrementally.

### [Seven Setters](https://github.com/vltpkg/vltpkg/blob/f98fabece724d6257a7186c807e9fefc0aeadf6d/src/gui/src/components/explorer-grid/selected-item/context.tsx#L389-L410)

```ts
  } finally {
    const {
      depLicenses,
      scannedDeps,
      averageScore,
      depWarnings,
      totalDepCount,
      duplicatedDeps,
      depFunding,
    } = counters


    setDepLicenses(depLicenses)
    setScannedDeps(scannedDeps)
    setDepsAverageScore(
      Math.floor((averageScore.score / averageScore.count) * 100),
    )
    setDepWarnings(Array.from(depWarnings.values()))
    setDepCount(totalDepCount || undefined)
    setDuplicatedDeps(duplicatedDeps)
    setDepFunding(depFunding)
  }
}
```

React does its best at batching state updates, but zustand updates synchronously. So we were forcing each component to "wake up" and check their state 7 times in a row within a single update.

### `lib/external-info.ts`

Previously, we were streaming updates directly into state as they came in. 

#### `selected-item/context`
```ts
for await (const d of fetchDetails(
            manifestHydratedSpec ?? finalSpec,
            abortController.signal,
            manifest ?? undefined,
          )) {
            store.setState(state => ({
              ...state,
              ...d,
            }))
          }
```

#### `lib/external-info`
```ts
while (true) {
    // when we reach the end of the promises queue, we have this extra logic
    // here to ensure that we yield all final results, including edge cases
    // such as when multiple promises fulfill at the same time.
    if (promisesQueue.length === 0) {
      let res: DetailsInfo = {}
      for (const p of await Promise.all(fulfilledPromises)) {
        for (const key of Object.keys(p)) {
          res = { ...res, [key]: p[key as keyof DetailsInfo] }
        }
      }
      yield res
      break
    }
    // yield results as soon as they're ready
    yield await Promise.race(promisesQueue)
  }
```

Therefore if `fetchDetails` yields 5 times (stars, licenses, manifest, readme etc) then the UI 'pops' 5 times, if on React `strictMode` this increases to 10

This is the module that got the biggest improvement. The previous logic was intense and hard to maintain with the callbacks, by leveraging [Effect](https://www.effect.website) we could break apart each meaningful operation into composable Effects, each operation (i.e. `fetchGithubStars`, `fetchGithubRepo`, `fetchReadme`, `fetchDownloadsLastYear`) could be modelled as an effect, with a wrapper for error tracking. By modelling each parallel fetch as an effect and combining them all, we could guarantee that the combined effect only succeeds once with a complete, aggregated data structure solving the layout thrashing issue by enforcing a single atomic update.

By introducing Effect, we have access to a type signature with a possible error / exception on the Effects, and then use them to display an 'indicator' in the UI when one or many fetches failed to happen.

And by removing the streaming we also remove the 'popcorning' and layout thrashing.

### DIFF
--- a/src/gui/src/lib/external-info.ts
+++ b/src/gui/src/lib/external-info.ts
@@ -1,106 +1,55 @@
-import { normalizeManifest } from '@vltpkg/types'
+import { Effect, Layer, Schema } from 'effect'
+import { FetchHttpClient, HttpClient } from '@effect/platform'
 import { compare, gt, prerelease } from '@vltpkg/semver'
+import { normalizeManifest } from '@vltpkg/types'
+import { parseReadme } from '@/lib/parse-readme.ts'
+import { parseAriaLabelFromSVG } from '@/utils/parse-aria-label-from-svg.ts'
 import {
   getRepoOrigin,
   getRepositoryApiUrl,
 } from '@/utils/get-repo-url.ts'
-import { parseReadme } from './parse-readme.ts'
+import { retrieveGravatar } from '@/utils/retrieve-gravatar.ts'
+
+// Re-export utilities
+export { retrieveGravatar as retrieveAvatar } from '@/utils/retrieve-gravatar.ts'
+export { parseAriaLabelFromSVG } from '@/utils/parse-aria-label-from-svg.ts'
+
 import type { Spec } from '@vltpkg/spec/browser'
-import type {
-  Repository,
-  Manifest,
-  NormalizedManifest,
-  Packument,
-  NormalizedContributorEntry,
-  Person,
-} from '@vltpkg/types'
+import type { NormalizedManifest, Repository } from '@vltpkg/types'
 
-export type Semver = `${number}.${number}.${number}`
-
-export const publicRegistry = 'https://registry.npmjs.org/'
-
-export const isSemver = (s: string): s is Semver =>
-  /^\d+\.\d+\.\d+$/.test(s)
-
-export const asSemver = (s: string): Semver => {
-  if (isSemver(s)) {
-    return s
-  }
-  throw new Error(`Invalid Semver: ${s}`)
-}
-
-export type GitHubRepo = {
-  owner?: {
-    avatar_url?: string
-    login?: string
-  }
-  updated_at?: string
-  stargazers_count?: number
-  organization?: {
-    login?: string
-  }
-  name?: string
-  default_branch?: string
-  commits_url?: string
-  contributors_url?: string
-}
-
-export type DownloadsRange = {
-  start: string
-  end: string
-  downloads: { downloads: number; day: string }[]
-}
-
-export type AuthorInfo = {
-  name: string
-  mail?: string
-  email?: string
-  url?: string
-  web?: string
-}
-
-export type ImageInfo = {
-  src: string
-  alt: string
-}
-
-export type Version = {
-  version: Semver
-  publishedDate?: string
-  publishedAuthor?: {
-    name?: string
-    email?: string
-    avatar?: string
-  }
-  unpackedSize?: number
-  integrity?: string
-  tarball?: string
-  gitHead?: string
-}
-
-export type Contributor = NormalizedContributorEntry & {
-  avatar?: string
-}
-
-export type DetailsInfo = {
-  author?: AuthorInfo
-  downloadsPerVersion?: Record<Semver, number>
-  downloadsLastYear?: DownloadsRange
-  favicon?: ImageInfo
-  publisher?: AuthorInfo
-  publisherAvatar?: ImageInfo
-  versions?: Version[]
-  greaterVersions?: Version[]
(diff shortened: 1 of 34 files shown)


src/gui/src/lib/external-info.ts
```diff
@@ -138,440 +87,1342 @@ export const readRepository = (
...
-              publisher: readAuthor(mani._npmUser),
-            }),
-          }
-        })
-        .catch(() => ({})),
+    // Use 'latest' if bareSpec is a range (contains ^, ~, >, <, =, etc.) or is missing
+    const isRange =
+      !spec.bareSpec ||
+      /[~^><=|]/.test(spec.bareSpec) ||
+      spec.bareSpec === '*'
```
ruyadorno: The `spec` object already has a ton of properties to identify tags vs ranges, etc

In this case you should check for `!!spec.range` but keep in mind there are more properties like `spec.distTag` that identify dist tags, - there are some references to the properties returned here: https://github.com/vltpkg/vltpkg/tree/main/src/spec#properties

src/gui/src/utils/get-repo-url.ts
```diff
@@ -25,6 +25,21 @@ export const normalizeUrl = (url: string): string => {
   return normalized
 }
 
+/**
+ * Strips common hosted git prefixes from repository strings.
+ * Handles formats like "github:owner/repo", "gitlab:owner/repo", etc.
+ */
+const stripHostedGitPrefix = (repo: string): string => {
+  // Common hosted git prefixes
+  const prefixes = ['github:', 'gitlab:', 'bitbucket:', 'gist:']
```
ruyadorno: Note that the list is actually here: https://github.com/vltpkg/vltpkg/blob/main/src/spec/src/browser.ts#L27-L32 in case you want to make this system a little more resilient to future changes in this list of hosts (which is unlikely to be changing often tbf) you could get it from:

```
import { defaultGitHosts } from '@vltpkg/spec/browser'
const prefixes = Object.keys(defaultGitHosts)
```

src/gui/src/utils/parse-aria-label-from-svg.ts
```diff
@@ -0,0 +1,12 @@
+export const parseAriaLabelFromSVG = (
+  svg: string,
+): string | undefined => {
+  const parser = new DOMParser()
+  const svgDoc = parser.parseFromString(svg, 'image/svg+xml')
+  const ariaLabel = svgDoc
+    .querySelector('svg')
+    ?.getAttribute('aria-label')
+  if (!ariaLabel) return undefined
+  const match = /[\d.]+\s*[kmb]?/i.exec(ariaLabel)
```
ruyadorno, line 26: is this necessary? always better if we can avoid regex 😅

ejkorol: I would agree with you as i also hate regex, though the parsed values in the aria label here could be very wonky and arbitrary like:

- `12`
- `12.5`
- `12.5k` or `12.5 K`
- `0.03 b`

I found this was easiest way to handle with them, but ill annotate the regex here in case we stumble upon this later

src/gui/src/lib/external-info.ts
ejkorol: This is a good catch and uncovered a small bug: `fetchRegistryManifest` is actually only called for external packages where we don't already have its manifest, and we can assume that its already been resolved to a semver value if its been published, so maybe we can just do:

```ts
url.pathname = `${spec.name}/${spec.semver || 'latest'}`
```

ruyadorno (approved):

PR #1334, opened 2025-12-08
tar: add node-tar inspired path sanitization

ruyadorno:
Include some of the sanitization from `tar` in our `@vltpkg/tar` unpack method.

### DIFF
--- a/src/tar/src/unpack.ts
+++ b/src/tar/src/unpack.ts
@@ -1,14 +1,13 @@
 import { error } from '@vltpkg/error-cause'
 import { randomBytes } from 'node:crypto'
 import { lstat, mkdir, rename, writeFile } from 'node:fs/promises'
-import { posix, resolve } from 'node:path'
+import { basename, dirname, resolve, sep } from 'node:path'
 import { rimraf } from 'rimraf'
 import { Header } from 'tar/header'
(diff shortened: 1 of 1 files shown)


evert (commented):
I think the best thing you can possibly do for security, is not to try to detect every possible way to escape the project root, but rather do something like:

```typescript
import { realpath } from 'node:path';

// Package root
const absoluteBasePath = realpath(packageBasePath);

// And then for each file/directory/symlink:
const itemAbsolutePath = path.join(absoluteBasePath, itemPath);

if (!itemAbsolutePath.startsWith(absoluteBasePath)) {
  throw new Error('no!');
}
```

This is very straightforward, and handles (as far as I know) all edge cases. No regexes. It might also be faster but don't know!

Your approach is similar to filtering for XSS in HTML input by using whitelists instead of building an AST and re-generating it. It's the classic 'Parse, don't validate!' advice.

I'm also curious why you're choosing to rewrite invalid paths instead of throwing. Wouldn't it be better to tell the user: this package tried to do something unsafe and we're refusing to continue, vs. quietly moving the file somewhere else?

evert (commented):
lgtm! solution is shorter/easier to read than the original =)

PR #1358, opened 2025-12-12
graph-run: add promise-level cycle detection

ruyadorno:
Track promise wait relationships to detect cycles before they cause
deadlocks in concurrent walks. When multiple entry points traverse
nodes that depend on each other, the existing node-level cycle check
may not catch the dependency in time.

- Add promiseWaiting map to Runner class
- Add #isWaitingFor for transitive promise cycle detection
- Check promise cycles in #step before adding to awaiting
- Clean up promiseWaiting on walk completion

### DIFF
--- a/src/graph-run/test/index.ts
+++ b/src/graph-run/test/index.ts
@@ -589,29 +589,65 @@ t.test('extreme async deadlock scenario', async t => {
   t.equal(visits[visits.length - 1], 'tap', 'visited tap last')
 })
 
+/**
+ * Helper to wait for a condition, with timeout protection.
+ * Loops with `await setTimeout(0)` until predicate returns true.
+ */
+const waitFor = async (
+  predicate: () => boolean,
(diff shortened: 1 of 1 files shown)


darcyclarke (approved):

src/graph-run/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "url": "git+https://github.com/vltpkg/vltpkg.git",
+    "directory": "src/graph-run"
+  },
+  "author": "vlt technology inc. <support@vlt.sh> (http://vlt.sh)",
+  "dependencies": {
+    "@vltpkg/error-cause": "workspace:*",
+    "@vltpkg/types": "workspace:*",
+    "polite-json": "catalog:",
+    "walk-up-path": "catalog:"
+  },
```
lukekarrys: I think these deps are not used and can be removed.

PR #1384, opened 2026-01-06
chore: update vlt-lock.json and fix linting errors

darcyclarke:
- updates outdated lockfile/deps
- cleans up linting errors


ruyadorno (dismissed):
All changes to `src/*` seems unrelated to the stated goal of the PR and should not be part of it. Same for the change in `scripts/rm-node-nodules.ts`.

lukekarrys:
The lines that are disabled for `@typescript-eslint/await-thenable` do not error when we install the deps with pnpm. I'm noting this because it's probably another bug with peer resolutions.

darcyclarke:
@lukekarrys _"when we install the deps with pnpm"_ did you test the same deps (ie. upgraded with a fresh lockfile in pnpm) as we have here in this branch for vlt? Just to confirm, we had to blow away the previous vlt lockfile which means we likely have differing deps installed across `vlt`/`pnpm` which may or may not mean the linting rules changed(?) although I guess that can be confirmed by pinning `@typescript-eslint`

lukekarrys:
I didn't blow away the pnpm lockfile but I did update all the linting related deps to their latest versions to test in https://github.com/vltpkg/vltpkg/pull/1387 to see if I got the same linting errors.

In this PR I also was able to remove the modifier and get everything working. What I think happened is that once the correct versions were in `node_modules` removing the modifier left those versions in place but corrected whatever was causing the erroneous `await-thenable` linting errors. We also might have been able to achieve the same thing with a more targeted modifier like `#@typescript-eslint #typescript`, but I think it's better to not have the modifier if we don't need it.

Issue #823, opened 2025-06-03
[DEV] update `scripts/check-deps` to use vlt query


lukekarrys:
We use `pnpm ls` currently to get a full list of all production dependencies in the `src/` workspace group. This should be updated to `vlt query`.

PR #1386, opened 2026-01-06
chore: use vlt to check deps

ruyadorno:
Updates the `scripts/check-deps.ts` script to use vlt internals.

Fixes: https://github.com/vltpkg/vltpkg/issues/823


lukekarrys:
This looks like the right idea but I think we should temporarily have the file support both pnpm and vlt by checking the existence of `node_modules/.pnpm` vs `node_modules/.vlt`. That way we can merge this before fully removing pnpm.

ruyadorno:
good point, I think I rather hold on landing this until the very end of the transition though

Issue #1378, opened 2025-12-19
[BUG] Inconsistent peer dependency resolution for pinned TypeScript version with typescript-eslint


lukekarrys:
### 📋 Bug Description

When a project has a **pinned** TypeScript version (e.g., `typescript@5.7.3`) and `typescript-eslint`, vlt inconsistently resolves the TypeScript peer dependency for different `@typescript-eslint/*` packages. Some packages get the pinned version (5.7.3) while others get a newer version (5.9.3), causing runtime errors.

### ✅ Expected Behavior

All `@typescript-eslint/*` packages should resolve their `typescript` peer dependency to the same version - the project's pinned `typescript@5.7.3`.

### ❌ Actual Behavior

vlt resolves peer dependencies inconsistently:
- `@typescript-eslint/parser` → 5.7.3 ✓
- `@typescript-eslint/eslint-plugin` → **5.9.3** ✗
- `@typescript-eslint/typescript-estree` → 5.7.3 ✓
- `@typescript-eslint/utils` → 5.7.3 ✓
- `@typescript-eslint/type-utils` → **5.9.3** ✗

This causes ESLint to fail with parsing errors due to AST incompatibilities between TypeScript versions.

### 🔄 Steps to Reproduce

1. Clone the reproduction: https://github.com/vltpkg/projects-that-error (or create a new project with the minimal `package.json` below)
2. Run:
```bash
cd projects/ts-eslint-peer-bug
vlt install
node --run lint
```

**Minimal `package.json`:**
```json
{
  "name": "ts-eslint-peer-bug",
  "type": "module",
  "devDependencies": {
    "eslint": "^9.39.0",
    "typescript": "5.7.3",
    "typescript-eslint": "^8.49.0"
  },
  "scripts": {
    "lint": "eslint ."
  }
}
```

### 🌍 Environment

- OS: darwin arm64
- Node: v22.18.0
- vlt: 1.0.0-rc.12

### 📜 Error Logs/Console Output

```shell
$ node --run lint

/path/to/src/index.ts
  0:0  error  Parsing error: Cannot use 'in' operator to search for 'illegalDecorators' in 42

✖ 1 problem (1 error, 0 warnings)


**Lockfile evidence (`vlt-lock.json`):**

"·npm·@typescript-eslint§eslint-plugin@8.50.0 typescript": "peer >=4.8.4 <6.0.0 ·npm·typescript@5.9.3"  <-- BUG
[...]

PR #1399, opened 2026-01-09
chore: update lockfile

ruyadorno:
Updating the lockfile for a fresh install after fixing some remaining peer dependency problems.

Refs: https://github.com/vltpkg/vltpkg/issues/1378


lukekarrys:
From looking at the PR locally, it seems like the latest changes created a lot more very similar peer contexts. Packages like react plugins are duplicated many more times than they were before. I'm not sure if this is intentional or not but it seems off to me.

I think we should test more and possibly make other fixes and try regenerating the lockfile again before merging this.

lukekarrys:
Also from looking at the CI run, the lockfile has a lot of churn after vlt install is run https://github.com/vltpkg/vltpkg/actions/runs/20867062039/job/59960676108#step:13:1

ruyadorno:
> Also from looking at the CI run, the lockfile has a lot of churn after vlt install is run https://github.com/vltpkg/vltpkg/actions/runs/20867062039/job/59960676108#step:13:1

I can replicate it on my local setup too, opened an issue here: https://github.com/vltpkg/vltpkg/issues/1404

ruyadorno:
Updated with a brand-new lockfile using the new DepID

darcyclarke (approved):

PR #1414, opened 2026-01-14
feat: add `vlt ping` command

darcyclarke:


src/cli-sdk/src/commands/ping.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+export const command: CommandFn<CommandResult> = async conf => {
+  const rc = new RegistryClient(conf.options)
+  const registryAlias = conf.positionals[0]
+
+  if (registryAlias) {
+    // Ping a specific registry by alias
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    const registries = conf.options.registries || {}
```
ruyadorno, line 102: Here you should use the `getOptions` method from `@vltpkg/spec` so that the end user can also point to the default named registries such as `npm`, `gh` and `jsr`, e.g:


```suggestion
    const { registries } = getOptions(conf.options)
```

Ref: https://github.com/vltpkg/vltpkg/blob/2a955d9fe8be2cb224083503e537c4274d87d7db/src/spec/src/browser.ts#L61-L91

src/cli-sdk/src/commands/ping.ts
```diff
@@ -0,0 +1,134 @@
...
+    const registry = registries[registryAlias]
```
ruyadorno, line 103: I believe we should merge both the `jsr-registries` and `registries` object here, so that we can also ping `jsr` (and any other custom jsr registries):


```suggestion
    const { 'jsr-registries': jsrRegistries } = getOptions(conf.options)
    const registry = ({
      ...jsrRegistries,
      ...registries,
    })[registryAlias]
```

src/cli-sdk/src/commands/ping.ts
```diff
@@ -0,0 +1,134 @@
...
+      const availableAliases = Object.keys(registries)
+      throw error('Unknown registry alias', {
+        found: registryAlias,
+        wanted:
+          availableAliases.length > 0 ? availableAliases : undefined,
+      })
+    }
+    return await pingRegistry(rc, registry, registryAlias)
+  } else {
+    // Ping all configured registries
```
ruyadorno, line 114: isn't it better to ping the default registry only by default? I would expect that to be the case 🤔

src/cli-sdk/src/commands/ping.ts
```diff
@@ -0,0 +1,134 @@
...
+    const registries = conf.options.registries || {}
+    const results: PingResult[] = []
+
+    // Always ping the default registry first
+    results.push(
+      await pingRegistry(rc, conf.options.registry, 'default'),
+    )
+
+    // Then ping all configured registry aliases
+    for (const [alias, registry] of Object.entries(registries)) {
```
ruyadorno, line 125: same as above, it would also need to look into `jsr-registries`

src/cli-sdk/src/commands/ping.ts
darcyclarke: I think it actually makes much more sense to ping all configured registries given that if they're configured there's the potential they'll be required during an install. If you want to scope/filter to a specific registry you can with the new optional arg (ex. `vlt ping npm`)

src/cli-sdk/src/commands/ping.ts
darcyclarke: I'd rather not hoist a non-command library's options (ie. `@vltpkg/spec`) & treat it as authoritative config; especially not in a way that would do that for a single command.

I noted this in our call before but I'll make a follow up PR for changes to ensure the `LoadedConfig` includes the missing registry values prior to ANY command being run. It sucks but it does likely mean duplicating what options `@vltpkg/spec` defined as internals back into `@vltpkg/config`.

src/cli-sdk/src/commands/ping.ts
darcyclarke: This is where I think there's an opportunity to improve the registry configuration. The delineation between "npm", "jsr" & "vlt" (in the future) is essentially metadata for the _type_ of API to expect or transform registry requests by. Bucketing at the top level feels like recreating the `dependency` vs. `devDependency` vs. `peerDependency` vs. `optionalDependency` nonsense.

If the type of registry is important then we should consider changing registry definitions to a be an object with a type to cleanly represent these definitions. Notably, this is also different then the config we should set to tell `vlt` how to treat/associate bare registry specs (ie. `react` vs `npm:react`).

ex. 
```json
{ 
  "default": "npm",
  "registry": "npm",
  "registries": {
    "npm": { 
      "url": "https://registry.npmjs.org", 
      "type": "npm" 
    },
    "jsr": { 
      "url": "https://npm.jsr.io", 
      "type": "jsr" 
    },
    "deno": { 
      "url": "https://deno.land", 
      "type": "jsr" 
    },
    "vlt": { 
      "url": "https://registry.vlt.io", 
      "type": "vlt" 
    },
    "msft": { 
      "url": "https://msft.artifactory.com", 
      "type": "npm" 
    },
    "vltpkg": { 
      "url": "https://vltpkg.vlt.io", 
      "type": "vlt" 
    }
  }
}
```

`--registry` could then accept both an `<alias>` (to a configured registry definition) &/or a `<url>` while independently defining what registry to associate with bare registry specifiers (ie. `--default=<alias|url>` or whatever name we want to give it).

This is a separate PR/work item though so I'll keep the changes as is for now which means JSR registries won't be included until they're consolidated with `"registries"` (ie. `vlt ping` only pings configured "npm" registries for now).

src/cli-sdk/src/commands/ping.ts
darcyclarke: Commented on this above as well

Issue #1404, opened 2026-01-12
[BUG] Lockfile churn on large monorepos


ruyadorno:
When installing large monorepos, we're observing random changes in the lockfile when installing from a clean slate vs using the lockfile.

PR #1419, opened 2026-01-16
graph: fix duplicated peer dep nodes

ruyadorno:
Removed unnecessary node duplication from the peer dep resolution implementation.

Fixes: https://github.com/vltpkg/vltpkg/issues/1404


ruyadorno:
CI is currently failing depending on the fix from https://github.com/vltpkg/vltpkg/pull/1420

Issue #96, opened 2024-07-10
[FEATURE] replace the bespoke `;` formatting with `URL`s for `DepID`s


darcyclarke:
### Description
Currently `DepID`s are a net-new string format that has no conceptual basis outside of `vlt` with the primary purpose being to encode the type of & resolved package id. This _may_ be fine if these were never to be seen by end-users but they are _very_ likely to bubble up during any kind of error handling/debugging & have quickly become fundamental identifiers used in various places in code/config (making it very likely to spill out into end-user output/documentation/config).

### Suggestion
Use URLs. Instead of creating a net-new format, leverage the existing `URL` format & use a basic lookup table to identify a `DepID`'s `type` based on it's `protocol`.

> [!IMPORTANT]
> Not all URLs would be valid/supported `DepID`s but **all** `DepID`s would be valid URLs - ex. `tel:`, `telnet:`, `mailto:`, `news:`, `urn:`  & `ldap:` aren't likely `type`s/sources for packages we'd ever support (although we should consider protecting against users configuring custom registry protocols which would collide with these & other [known/common URI schemes](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml)).

### Examples

> [!NOTE]
> For brevity/clarity, values are not percent encoded but all `DepID`s should likely take the form post `encodeURIComponent` so as to be valid directory names.
 
| Before | After |
| --- | --- |
| `registry;;foo@1.0.0` or `;;foo@1.0.0`, if #89 happens | `npm:foo@1.0.0` |
| `registry;vlt;x@1.2.3` | `vlt:x@1.2.3` |
| `file;./x.tgz` | `file:./x.tgz` |
| `git;git+ssh://host.com/x.git` | `git+ssh://host.com/x.git` |
| `git;github:a/b;branch` | `github:a/b#branch` |
| `remote;https://x.com/x.tgz` | `https://x.com/x.tgz` |
| `workspace;a` | `workspace:a` |

### Implementation

An example `types`/`protocols` lookup table seems pretty straightforward (can change/expand as needed):

```ts
const types = {
  npm: 'registry',
  git: 'git',
  'git+ssh': 'git',
  'git+http': 'git',
  'git+https': 'git',
  'git+ftp': 'git',
  'git+file': 'git',
  'git+rsync': 'git',
  github: 'git',
  bitbucket: 'git',
  gitlab: 'git',
  http: 'remote',
  https: 'remote',
  file: 'file',
  workspace: 'workspace'
[...]

PR #1420, opened 2026-01-19
dep-id: replace delimiters

ruyadorno:
Replaces the `@vltpkg/dep-id` encode / decode implementation to
avoid using `encodeURIComponent` and `decodeURIComponent` and
instead use a custom encode system.

Also removes usage of the special utf8 characters in DepIDs.

This avoids issues in which external tools fail to load specific
files from a vlt-cli install due to usage of loaders that are
automatically decoding URI-encoded strings and fails to read
paths using special characters.

Refs: https://github.com/vltpkg/vltpkg/issues/96
Refs: https://github.com/vltpkg/vltpkg/issues/1404

### DIFF
--- a/.cursor/rules/graph/lockfiles.mdc
+++ b/.cursor/rules/graph/lockfiles.mdc
@@ -280,7 +280,7 @@ examples:
   - input: |
       # Working with lockfile data format (including new fields)
       const lockfileData = {
-        lockfileVersion: 0,
+        lockfileVersion: 1,
         options: { registry: 'https://registry.npmjs.org/' },
         nodes: {
           'registry:foo@1.0.0': [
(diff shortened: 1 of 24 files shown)


src/cache/README.md
```diff
@@ -5,7 +5,7 @@
 The filesystem cache for `@vltpkg/registry-client`, but also, a
 general-purpose filesystem-backed [LRUCache](http://npm.im/lru-cache)
 
-**[Usage](#usage)** · **[Note](#note)**
+**[Usage](#usage)** ~ **[Note](#note)**
```
ruyadorno:
```suggestion
**[Usage](#usage)** · **[Note](#note)**
```

ruyadorno:
vlt CI actions are going to be failing until the lockfile is updated

PR #1430, opened 2026-01-22
chore: remove vlt serve cmd

ruyadorno:
Removes the browser-based UI along with
the `vlt serve` vsr shortcut.


src/cli-sdk/src/commands/list.ts
```diff
@@ -6,12 +6,10 @@ import {
   mermaidOutput,
   GraphModifier,
 } from '@vltpkg/graph'
-import LZString from 'lz-string'
```
lukekarrys, line 9: Can we remove `lz-string` from the package.json too?

PR #1434, opened 2026-01-28
chore: use relative value for parallelism

ruyadorno:
Instead of arbitrarily multiplying the amount of parallel operations by 8 (which is likely the most common number of thread we've been testing against) let's use the os-provided thread number as the multiplying factor.


ruyadorno:
it looks like the performance is not as great on CI / benchmark machines, I'm going to hold off on this one for now

PR #1482, opened 2026-02-19
fix(graph): pass actual graph to idealBuild in uninstall for manifest hydration

### LINKED ISSUE #1426: [BUG] `vlt uninstall` fails to remove dependency
In a very simplistic install, once there's more than one dependency running `vlt uninstall <pkg>` fails to update the hidden lockfile after removing the dependency, e.g:

```
$ vlt init
...
$ vlt i which
resolving dependencies ✓ > extracting files ✓
4 cache hits
Done in 323ms
$ vlt query '*'
test-11
└─┬ which@6.0.0
  └── isexe@3.1.1

$ vlt i abbrev
resolving dependencies ✓ > extracting files ✓
3 cache hits
Done in 231ms
$ vlt query '*'
test-11
├── abbrev@4.0.0
└─┬ which@6.0.0
  └── isexe@3.1.1

$ vlt uninstall abbrev
resolving dependencies ✓ > extracting files ✓
Done in 231ms
$ vlt query '*'
test-11
├── abbrev@4.0.0
└─┬ which@6.0.0
  └── isexe@3.1.1

# The dependency has been removed from lockfile and node_modules folder:
cat vlt-lock.json
{
  "lockfileVersion": 1,
  "options": {},
  "nodes": {
    "~npm~which@6.0.0": [0,"which","sha512-f+gEpIKMR9faW/JgAgPK1D7mekkFoqbmiwvNzuhsHetni20QSgzg9Vhn0g2JSJkkfehQnqdUAx7/e15qS1lPxg==",null,null,null,null,null,{  "node-which": "bin/which.js"}]
  },
  "edges": {
    "file~_d which": "prod ^6.0.0 ~npm~which@6.0.0"
  }
}
$ tree node_modules/.vlt
node_modules/.vlt
├── node_modules
│   └── which -> ../~npm~which@6.0.0/node_modules/which
└── ~npm~which@6.0.0
    └── node_modules
        └── which
$ cat node_modules/.vlt-lock.json | jq -r '.nodes | keys[]'
~npm~abbrev@4.0.0
~npm~isexe@3.1.1
~npm~which@6.0.0
```

baudbot-agent:
## Summary

Fixes #1426 — `vlt uninstall` fails to update the hidden lockfile (`node_modules/.vlt-lock.json`), leaving stale entries behind even after a dependency is correctly removed from the main lockfile and `node_modules`.

## Root Cause

In `uninstall.ts`, the actual graph was loaded **after** `idealBuild` and was never passed to it. This meant `idealBuild` loaded the virtual graph from `vlt-lock.json` without an `actual` reference.

Since `vlt-lock.json` does not store manifest data, nodes loaded from it were missing manifests. When `reify` later called `saveHidden` with `throwOnMissingManifest: true`, `lockfileData()` threw an error for those manifest-less nodes. The empty `catch {}` silently swallowed the error, so the hidden lockfile was never written — leaving stale entries behind.

By contrast, `install.ts` already loads the actual graph **before** `idealBuild` and passes it via the `actual` option. This allows `loadNodes` (in `lockfile/load-nodes.ts`) to hydrate missing manifests from the actual graph's installed packages (`const mani = manifest ?? referenceNode?.manifest`).

## Fix

- Load the actual graph **before** `idealBuild` in `uninstall.ts`
- Pass it to `idealBuild` via the `actual` option — matching the pattern used by `install.ts`
- Updated the existing unit test to verify `actual` is passed to `idealBuild`
- Updated the snapshot to reflect the corrected ordering

## Test

```
cd src/graph && vlr test -Rtap --disable-coverage test/uninstall.ts
# 12/12 pass
```


ruyadorno:
@cursoragent let's address the following: `modifiers` is loaded in `src/graph/src/uninstall.ts` but not passed into `idealBuild` or later `reify`. Unlike `src/graph/src/install.ts`/`src/graph/src/update.ts`, this means uninstall won’t apply `vlt.json` modifiers during the ideal build and the hidden lockfile saved during reify won’t include the modifiers config in its lockfile options. Pass `modifiers` into `idealBuild` and through to `reify` to keep behavior consistent across commands.

Update the PR with the changes and make sure to run the complete validation described at `.cursor/rules/code-validation-workflow.mdc`

PR #1507, opened 2026-03-02
fix(registry-client): handle 412 Precondition Failed on revalidation

darcyclarke:
Some registries (notably npm.jsr.io) return HTTP 412 when they receive conditional headers (If-None-Match / If-Modified-Since) on GET requests, instead of the expected 304 or 200. Handle this by stripping the conditional headers and retrying the request.

### DIFF
--- a/www/docs/astro.config.mts
+++ b/www/docs/astro.config.mts
@@ -93,6 +93,11 @@ export default defineConfig({
           autogenerate: { directory: 'cli' },
         },
         {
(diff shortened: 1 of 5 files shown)


evert:
I wonder if it's worth reporting upstream as well.

Regardless I think you can treat `412` as a `304`, because it's basically saying: "I don't need to fulfill this request because the state has not changed".

PR #1510, opened 2026-03-02
docs: add migration guides for npm, yarn & pnpm

### LINKED ISSUE #97: [DOCS] Migrating from `pnpm`, `npm`, `yarn` & `bun`
### Description

Making it easy for end-users to understand how they can migrate projects from other package managers. 

### Suggestion
- make code changes to support &/or help migrate tools (ex. don't allow multiple types of known lockfiles to exist, do one-time configuration/lockfile migration on behalf of users)
- create quick-starters &/or migration guide docs

darcyclarke:
Closes #97

### DIFF
--- a/www/docs/astro.config.mts
+++ b/www/docs/astro.config.mts
@@ -93,6 +93,11 @@ export default defineConfig({
           autogenerate: { directory: 'cli' },
         },
         {
+          label: 'Migration',
+          collapsed: true,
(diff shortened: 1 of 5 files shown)


ruyadorno:
I believe removing `node_modules` might be a required step (that is not currently being called out in these docs) in all these migrations.

PR #1537, opened 2026-03-06
chore: clean up workspaces/repo

darcyclarke:
- removed `src/gui` (`vlt serve` was already removed in a previous PR)
- removed `src/server` (dependency of the gui)
- removed `src/vsr` (consumer of the gui)
- cleaned up `infra` (which had a bunch of gui & friends config), particularly the compiled/variants publishing (can reintroduce shipping compiled variants in future releases)

### DIFF
--- a/infra/smoke-test/README.md
+++ b/infra/smoke-test/README.md
@@ -41,10 +41,7 @@ appears in the lockfile:
 import t from 'tap'
 import { readFileSync } from 'node:fs'
 import { join } from 'node:path'
-import {
-  runMultiple,
-  allVariants,
-} from './fixtures/run.ts'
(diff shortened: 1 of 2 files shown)


lukekarrys:
With this change we are only using 3 functions exported from `src/server` and they are only used in `src/cli-sdk`. I would recommend either removing the unused functionality from `src/server` OR moving the used functions to `src/cli-sdk` and then deleting the `src/server` workspace.

darcyclarke:
@lukekarrys agreed - I actually made that change in another branch that I dropped. Will add those changes here as `src/server` is highly relevant/tied to the GUI.

PR #1542, opened 2026-03-13
feat: implement skills support for vlt

### LINKED ISSUE #1540: [FEATURE] Support skills
### 📋 Description

During install - keep the list of known locations of `skills.md` files & then symlink them (similar to how we handle bins) into the project's root `./skills/<package name>/...` folder

We should add a net-new flag to enable/disable installing these similar to how we did with scripts (ie. `--allow-scripts` & `vlt build`). The great thing there is that we can utilize our selector syntax for fine-grained allow lists (although I'm sure many people will just want to do `:not(*)` or `*` in a lot of cases).

The net-new `install` flag/option should be something like `--allow-skills` & this should really be passed down & be part of a net-new command (like `vlt build`) under-the-hood (ex. `vlt skills`) since that's where we'd do the lookup/linking for the skills themselves (which is different then script/bin linking although there may be overlap with `vlt build` - we should consider if how these to relate or if there's net-new capabilities we'd want to add to a dedicated `vlt skills` command beyond just mounting & managing skills like we do with package bins).

### Considerations
- we already have the framework for managing dependencies this feels like a natural win/extension of what dependencies will be shipping now & moving forward (ie. many existing packages have begun & will begin distributing these)
- we could try to promote a standard way of defining skills in `package.json` but that seems like extra overhead compared to how people are currently defining these (ie. utilizing the filesystem) & we can easily extract this understanding during install (given that we're placing these files on the filesystem to begin with)

### 🔍 Research & Prior Art

- https://skills.sh/ & https://github.com/vercel-labs/skills exists but only supports github repos - it also has no versioning/lockfiles so we get that for free here by being a package manager that can mount these files on your behalf

vltbaudbot:
Implements skills support as described in #1540

## What this PR does

This PR implements skills support for the vlt package manager, adding:

1. **`vlt skills` command** - A new CLI command that can:
   - List discovered skills (`vlt skills list` or `vlt skills`)
   - Link skills to project directory (`vlt skills link`)
   - Unlink skills (`vlt skills unlink`)

2. **`--allow-skills` config option** - Similar to `--allow-scripts`:
   - Uses DSS query syntax for allow/deny lists
   - Patterns like `*` (allow all), `:not(*)` (deny all), or specific selectors
   - When set, install automatically links skills after reify

3. **Integration with install** - After reify completes, if `--allow-skills` is set, discovers and symlinks skills from installed packages

## How it works

- **Discovery**: Scans packages for skill files: `skills.md`, `SKILL.md`, `*.skill.md`
- **Linking**: Creates symlinks in `./skills/<package-name>/` pointing to skill files
- **Filtering**: Uses existing DSS query infrastructure for package selection
- **Safety**: Non-blocking - skills linking failures don't break installs

## Implementation details

- Added `skills.ts` module in `@vltpkg/graph` with discovery and linking logic
- Created `skills` command in CLI SDK following `build` command patterns
- Integrated with existing bin-linking and reify infrastructure
- Added comprehensive tests for both CLI and graph functionality

## Examples

```bash
# Install with skills linking enabled for all packages
vlt install --allow-skills=*

# List discovered skills
vlt skills list

# Link skills from specific packages
vlt skills link --target="#express, #lodash"

# Remove skill links
vlt skills unlink
```

Closes #1540


usrrname:
#1679  is being implemented with this in mind

PR #1609, opened 2026-05-02
fix(cli-sdk): honor -- as terminator in exec trailing option extraction

### LINKED ISSUE #1608: vlt run/exec intercepts flags from script args (regression from #1531)
## Summary

PR #1531 made several vlt config flags position-independent in `vlt run` / `vlt exec` — they can now appear after the script name. The change is reasonable, but there's no escape hatch: a script that takes one of these flag names can no longer receive it as an arg, and `--` does not terminate the extraction.

## Source

The new logic lives in `#extractTrailingOptions()` at `src/cli-sdk/src/exec-command.ts:165-220`. After the jackspeak parser captures everything past the command name as positionals (via `stopAtPositionalTest` in `src/cli-sdk/src/config/definition.ts:144-155`), this method re-scans those positionals for a hardcoded list of vlt options and pulls them back into `conf.values`.

## Hardcoded list of intercepted flags

From `exec-command.ts:141-158`:

- `--scope`
- `--workspace` / `-w`
- `--workspace-group`
- `--recursive` / `-R`
- `--if-present`
- `--bail`
- `--script-shell`
- `--view`

Any script that accepts a flag with the same name will never see that flag.

## Reproduction

Using a trivial script that prints `process.argv.slice(2)` (named `print-args` here):

| Command | Script receives | Notes |
|---|---|---|
| `vlr --bail print-args` | `[]` | baseline — `--bail` before script consumed as vlt config (correct) |
| `vlr print-args --bail` | `[]` | trailing `--bail` intercepted (the new #1531 behavior — fine in itself) |
| `vlr print-args -- --bail` | `['--']` | **the problem**: `--` is not honored as a terminator, so there is no way to forward `--bail` to the script |

## Proposed fix

Honor `--` as a terminator inside `#extractTrailingOptions()`. Once `--` is encountered in the trailing args, stop scanning and pass the remainder through to the script unchanged (and drop the `--` itself, per POSIX convention).

While in there, the comment at [src/cli-sdk/src/config/definition.ts#L147-L150](https://github.com/vltpkg/vltpkg/blob/f8ebf2d30371d4f9bea060daff9f69abe2b2dffe/src/cli-sdk/src/config/definition.ts#L147-L150) still describes pre-#1531 behavior and should be updated to reflect that vlt config flags can now appear on either side of the command name.

vltbaudbot:
## Summary

PR #1531 made vlt config flags position-independent in `vlt run` / `vlt exec` — they can appear after the script name. But there was no escape hatch: `--` didn't terminate the extraction, so scripts that accept flags sharing names with vlt config flags (like `--bail`) could never receive them.

### `src/cli-sdk/src/exec-command.ts`
- **`#extractTrailingOptions()`**: When a bare `--` is encountered, stop scanning for vlt options. Everything after `--` is passed through to the script unchanged (the `--` itself is consumed, per POSIX convention).

### `src/cli-sdk/src/config/definition.ts`
- Updated the `stopAtPositionalTest` comment to reflect the current behavior (post-#1531) and document the `--` terminator.

### Tests
- Added 3 unit tests in `test/exec-command.ts` for direct `--` terminator behavior
- Added 3 integration tests in `test/commands/run.ts`:
  - `--` forwards known vlt flags to script
  - `--` forwards multiple flags to script
  - vlt options before `--` are extracted, flags after are forwarded

## Reproduction

```bash
# Before: --bail is intercepted by vlt
vlt run test -- --bail   # script receives ['--'], --bail consumed

# After: -- terminates extraction
vlt run test -- --bail   # script receives ['--bail']
```

Closes #1608

Co-authored-by: Luke Karrys <luke@lukekarrys.com>

### DIFF
--- a/src/cli-sdk/src/config/definition.ts
+++ b/src/cli-sdk/src/config/definition.ts
@@ -145,10 +145,15 @@ const j = jack({
   stopAtPositionalTest: arg => {
     if (stopParsing) return true
     const a = arg as keyof Commands
-    // we stop parsing AFTER the thing, so you can do
-    // vlt run --vlt --configs scriptName --args --for --script
-    // or
-    // vlt exec --vlt --configs command --args --for --command
+    // We stop parsing AFTER the command name, so options that
+    // appear after the script/command name end up as positionals.
+    // ExecCommand.#extractTrailingOptions() then re-scans those
+    // positionals for known vlt options (--scope, --workspace,
+    // --recursive, etc.) regardless of position.  A bare `--`
+    // terminates that re-scan so flags after it are always
+    // forwarded to the script:
+    //   vlt run scriptName --scope "..." --bail  ← vlt gets both
+    //   vlt run scriptName -- --bail              ← script gets --bail
     if (stopParsingCommands.includes(commands[a])) {
       stopParsing = true
     }
--- a/src/cli-sdk/src/exec-command.ts
+++ b/src/cli-sdk/src/exec-command.ts
@@ -161,6 +161,10 @@ export class ExecCommand<B extends RunnerBG, F extends RunnerFG> {
    * Scan trailing args for known vlt options that ended up as
    * positionals due to stop-at-positional parsing. Extract them
    * and apply to conf.values so they're available via conf.get().
+   *
+   * A bare `--` terminates extraction: everything after it is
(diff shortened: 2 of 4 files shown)


lukekarrys:
This is no longer needed after #1619

PR #1660, opened 2026-06-15
chore(ci): update malware check queries in CI workflow

darcyclarke:
This pull request makes a minor update to the malware checking step in the CI workflow. The change ensures that only critical malware findings are checked, rather than all malware detections.

- CI workflow update:
  * Changed the malware check in `.github/workflows/ci.yml` to use `:malware(critical)` instead of `:malware`, focusing the check on "critical" findings only.

A follow up issue to correct the root cause has been created here (& we'll need to revert this once a newer version lands with a fix to the query selector): https://github.com/vltpkg/vltpkg/issues/1661


evert:
Thank you!

PR #1699, opened 2026-08-03
feat(query): return informational messages for empty results in default human view

usrrname:
Add informational stdout messages when `vlt query` returns no results:

## Changes

- If scope query matches nothing: `No packages found matching scope query: <query>`
- if workspace filter matches nothing: `No matching workspaces found: <workspace>`
- if main query matches nothing: `No packages found matching query: <query>`

Added unit tests in `src/cli-sdk/test/commands/query.ts` for:
1. scope query matches nothing — verifies stdout is called with "No packages found matching scope query:" when scope selector matches no packages
2. workspace filter matches nothing — verifies stdout is called with "No matching workspaces found:" when workspace filter doesn't match any workspace
3. main query matches nothing (with graph) — verifies stdout is called with "No packages found matching query:" when main query matches no packages (graph defined)
4. expect-results overrides empty result message — verifies `--expect-results` takes priority and throws validation error instead of logging empty message

Previously these scenarios silently returned empty output with no user feedback. 
Messages are only shown when a graph is loaded (not when resolving via host contexts without a local `package.json`).

I primarily became concerned about this as I plan on promoting the `vlt query :malware` command and when I run it in a workspace where no query results are found for that command, I get no feedback. 

Before:
Screenshot 2026-08-03 at 9 31 09 AM

After:
Screenshot 2026-08-03 at 9 32 01 AM

## Testing this locally

```bash
node src/cli/dist/index.js query ':malware'

# Or run the test suite:
cd ~/vltpkg/src/cli-sdk
vlx tap test/commands/query.ts

To test the specific scenarios manually, create a temp directory with a package.json and run from there:
mkdir /tmp/test-query && cd /tmp/test-query
echo '{"name":"test","version":"1.0.0","dependencies":{"foo":"^1.0.0"}}' > package.json
vlt install
~/vltpkg/src/cli/dist/index.js query ':malware'
~/vltpkg/src/cli/dist/index.js query '#nonexistent'
~/vltpkg/src/cli/dist/index.js query --scope='#nonexistent' '*'
```

### DIFF
--- a/src/cli-sdk/src/commands/query.ts
+++ b/src/cli-sdk/src/commands/query.ts
@@ -202,10 +202,12 @@ export const command: CommandFn<QueryResult> = async conf => {
     )
     scopeNodes = resultNodes
     if (scopeNodes.length === 0) {
-      stdout(
-        'No packages found matching scope query:',
-        scopeQueryString,
-      )
+      if (conf.values.view === 'human') {
+        stdout(
+          'No packages found matching scope query:',
+          scopeQueryString,
+        )
+      }
       return {
         importers: new Set(),
         edges: [],
@@ -234,7 +236,9 @@ export const command: CommandFn<QueryResult> = async conf => {
       }
     }
     if (importers.size === 0) {
-      stdout('No matching workspaces found:', conf.values.workspace)
+      if (conf.values.view === 'human') {
+        stdout('No matching workspaces found:', conf.values.workspace)
+      }
       return {
         importers: new Set(),
         edges: [],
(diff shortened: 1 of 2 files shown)


ruyadorno (changes requested):
This message should only be displayed for the default (human) output.

### ❌ Current

```
$ vlt query ':score(100)'
No packages found matching query: :score(100)

$ vlt query ':score(100)' --view=json
No packages found matching query: :score(100)
[]
$ vlt query ':score(100)' --view=mermaid
No packages found matching query: :score(100)
flowchart TD

$ vlt query ':score(100)' --view=count
No packages found matching query: :score(100)
0
```

### ✅ Expected

```
$ vlt query ':score(100)'
No packages found matching query: :score(100)

$ vlt query ':score(100)' --view=json
[]
$ vlt query ':score(100)' --view=mermaid
flowchart TD

$ vlt query ':score(100)' --view=count
0
```

ruyadorno (approved):

PR #1728, opened 2026-08-05
fix: redirect /migration/from-pnpm to /cli/migration/from-pnpm

usrrname:
## Problem
Old URL `/migration/from-pnpm` needed to redirect to the new location at `/cli/migration/from-pnpm` after content reorganization resulted in 404 https://vltpkg.slack.com/archives/C0BN0KQJ86M/p1785876626436299

## Solution
Implemented a hybrid redirect approach optimized for static Astro sites on Vercel:

1. **Vercel platform-level 301 redirects** (`vercel.json`)
   - Returns proper HTTP 301 status codes
   - Better for SEO
   - Handles at edge before static files

2. **Static HTML fallback** (`www/docs/public/migration/from-pnpm.html`)
   - Ensures [most popular] redirect works even if platform redirect doesn't apply
   - Includes meta refresh + JavaScript redirect
   - User-friendly message

3. **Fixed typedoc.mjs git handling**
   - Gracefully handles missing git in Vercel build environment
   - Prevents build failures during deployment

## Changes
- Added `vercel.json` with outputDirectory and 301 redirects for migration URLs
- Created static HTML redirect file with fallback methods
- Fixed typedoc.mjs error handling for Vercel builds
- Removed Vercel adapter (not needed for static site)

## Notes
- Tested multiple approaches (Astro middleware, route files, config redirects)
- Static Astro sites on Vercel have limited redirect options
- This hybrid approach provides both SEO-friendly 301s and reliable fallback

Expected behaviour: 

https://docs-git-fix-docs-redirect-vlt.vercel.app/migration/from-pnpm  redirects to  https://docs-git-fix-docs-redirect-vlt.vercel.app/cli/migration/from-pnpm

https://docs-git-fix-docs-redirect-vlt.vercel.app/migration/from-yarn redirects to  https://docs-git-fix-docs-redirect-vlt.vercel.app/cli/migration/from-yarn

https://docs-git-fix-docs-redirect-vlt.vercel.app/migration/from-npm  redirects to https://docs-git-fix-docs-redirect-vlt.vercel.app/cli/migration/from-npm


www/docs/typedoc.mjs
```diff
@@ -16,10 +16,26 @@ import { fileURLToPath } from 'node:url'
...
+      try {
+        execSync('git init')
+        execSync(`git remote add origin ${remote}`)
+      } catch {
+        // if git operations fail, continue — typedoc will use fallback for source links
+      }
+    }
+  } catch {
+    // silently ignore any git setup errors
+  }
```
ruyadorno: Sounds like a good idea to at least print this error here.

```suggestion
  } catch (err) {
    console.error(err)
  }
```

usrrname: Oh yeah great catch!

www/docs/public/migration/from-pnpm.html
```diff
@@ -0,0 +1,72 @@
...
+      // Fallback redirect if meta refresh doesn't work
+      if (
+        window.location.pathname === '/migration/from-pnpm' ||
+        window.location.pathname === '/migration/from-pnpm/'
+      ) {
+        window.location.replace('/cli/migration/from-pnpm')
+      }
+    </script>
+  </body>
+</html>
```
lukekarrys: I think without the adapter the server side vercel.json redirects work and we dont need a separate client-side one? Using curl against the deploy preview I get the correct 308:

```
╰❯ curl -sS -D - -o /dev/null \
         https://docs-git-fix-docs-redirect-vlt.vercel.app/migration/from-pnpm
HTTP/2 308
cache-control: public, max-age=0, must-revalidate
content-type: text/plain
date: Wed, 05 Aug 2026 18:32:01 GMT
location: /cli/migration/from-pnpm
refresh: 0;url=/cli/migration/from-pnpm
server: Vercel
strict-transport-security: max-age=63072000; includeSubDomains; preload
x-robots-tag: noindex
x-vercel-id: sfo1::pxn4w-1785954721869-a60df83bee8d
```

www/docs/astro.config.mts
```diff
@@ -3,7 +3,6 @@ import starlight from '@astrojs/starlight'
 import react from '@astrojs/react'
 import sitemap from '@astrojs/sitemap'
 import tailwind from '@astrojs/tailwind'
-import vercel from '@astrojs/vercel'
```
lukekarrys, line 6: If this isnt needed anymore, can we remove from the package.json and lockfile?

www/docs/typedoc.mjs
```diff
@@ -16,10 +16,26 @@ import { fileURLToPath } from 'node:url'
...
+      try {
+        execSync('git init')
+        execSync(`git remote add origin ${remote}`)
+      } catch {
+        // if git operations fail, continue — typedoc will use fallback for source links
+      }
+    }
+  } catch (err) {
+    console.error(err)
+  }
```
lukekarrys, line 38: Was this fixing a previously broken build? And I think we probably do want to continue to break the build in Vercel envs if git is not available. Otherwise we can end up with a lot of broken links in the generated typedoc pages.

lukekarrys (changes requested):
I tested the preview deployment and I'm getting correct redirects now. But a few changes and things I see to make this simpler.

www/docs/typedoc.mjs
usrrname: Anytime I pushed a new change to the branch, the build would break on vercel (I imagine maybe build cache would run?), so I was running the build locally and found this error would happen. if it's not needed, we can scrap it

www/docs/astro.config.mts
usrrname: This isn't part of this PR anymore... it was a previous commit. We're not using the vercel adapter here

www/docs/public/migration/from-pnpm.html
usrrname: yes I added this in case I was crazy as a fallback since trying the first 3 most commonly suggested things in docs out there didn't work.

usrrname: This is a client side fallback just in case

usrrname: Got rid of it!

www/docs/astro.config.mts
lukekarrys: I think this was the only place it was being used. So I think we can drop it from our dep graph entirely. Can you also remove it from the docs package.json and run `vlt i` to update the lockfile and see if that still works? https://github.com/vltpkg/vltpkg/blob/32fb0a10834fcf327ab6598dbabab8995ea98b7e/www/docs/package.json#L19

www/docs/typedoc.mjs
lukekarrys: ok that makes sense. I spot checked the preview for the typedoc links and they all worked.

lukekarrys (approved):
Looks good to me! We can drop the dep and update the lockfile in a follow up if we need.

PR #1730, opened 2026-08-06
graph: better node name validation

ruyadorno:
Adds additional validation and sanitization to guard against usage of accidental path segments in a package.json name, which leads to broken installs. Also added the same sanitization logic to bin paths.


src/graph/src/lockfile/load-nodes.ts
```diff
@@ -25,6 +29,9 @@ export const loadNodes = (
     nodeCount > 50 ? new Map<string, string>() : null
 
   for (const [id, lockfileNode] of entries) {
+    // lockfile keys have never been through `encode`, so they can hold
+    // anything, and the id is the store directory name
+    asDepID(id)
```
lukekarrys, line 34: [P1] Validate lockfile-supplied node names

The new validation checks only the DepID key. A remote or git node such as `[0, '../../../../../../outside']` still passes its unsafe name to `Graph.addNode`; the resulting default `Node.location` resolves outside the project during extraction. Validate `name` or `manifest.name` before constructing the node.

src/graph/src/reify/index.ts
```diff
@@ -216,6 +217,10 @@ const reify_ = async (
   // any failed/removed optional deps are not reflected in the lockfile
   // data as it is saved.
   const lfData = lockfileData(options)
+
+  // never write through a symlinked store
+  await ensureStoreDirs(scurry, remover)
```
lukekarrys: [P1] Guard the no-change lockfile write path

`ensureStoreDirs()` is only called inside `reify_()`, but the no-change branch returns earlier and may call `saveHidden()`. For an empty project whose `node_modules` is a symlink and whose hidden lockfile is missing, this writes through the symlink. Run the store-directory check before any early-return writes.

src/graph/src/reify/ensure-store-dirs.ts
```diff
@@ -0,0 +1,23 @@
...
+ * are symlinks the writes land wherever they point. Remove anything in
+ * their place that is not a directory. One syscall per distinct dir.
+ */
+export const ensureStoreDirs = async (
+  scurry: PathScurry,
+  remover: RollbackRemove,
+): Promise<void> => {
+  for (const dir of ['node_modules', 'node_modules/.vlt']) {
+    const path = scurry.resolve(dir)
+    if (checked.has(path)) continue
```
lukekarrys: [P1] Recheck store directories on every reification

This process-global cache permanently trusts a path after its first check, including when the path did not yet exist. If a later reification in the same process finds that directory replaced by a symlink, it skips `lstat()` and writes through it. A security boundary must recheck each invocation; remove this cache or scope it to one operation.

src/dep-id/src/browser.ts
```diff
@@ -461,10 +470,9 @@ export const hydrateTuple = (
       if (!first) {
         throw error('no name/path on workspace id', { found: tuple })
       }
-      res =
-        name && name !== first ?
-          Spec.parse(name, `workspace:${first}@*`, options)
-        : Spec.parse(first, `workspace:*`, options)
+      const workspaceName =
+        name ?? first.slice(first.lastIndexOf('/') + 1)
+      res = Spec.parse(workspaceName, 'workspace:*', options)
```
lukekarrys: [P2] Preserve the workspace path when hydrating

Replacing the explicit workspace spec with `workspace:*` discards the path stored in the DepID. For example, hydrating `workspace~packages+real` under alias `alias` now sets `spec.workspace` to `alias`; `satisfies()` then compares `monorepo.get('packages/real')` with `monorepo.get('alias')` and rejects a valid path-targeted workspace. Keep `first` in a `workspace:${first}@*` bare spec while using the safe alias as the package name.

src/workspaces/src/index.ts
```diff
@@ -697,7 +697,7 @@ export class Workspace {
     this.path = path
     this.fullpath = fullpath
     this.manifest = manifest
-    this.name = manifest.name ?? path
+    this.name = manifest.name ?? basename(path)
```
lukekarrys, line 700: [P2] Keep unnamed workspace fallbacks unique

Using only `basename(path)` makes distinct unnamed workspaces such as `packages/demo` and `examples/demo` both become `demo`; `#loadWS()` then throws `Duplicate workspace name found`, although unnamed manifests and both unique paths were previously supported. Use a path-derived filesystem-safe fallback that remains unique.

lukekarrys (commented):
Code review: five actionable findings, attached inline.

src/graph/src/lockfile/load-nodes.ts
ruyadorno: the lockfile is surface the user owns, so it's counterproductive to be overly asserting these values here

src/dep-id/src/browser.ts
ruyadorno: this looks like a good catch, will give it a try

src/workspaces/src/index.ts
ruyadorno: this change aligns the client with how the original npm works behind the scenes when a name can't be found in the `package.json` file

src/dep-id/src/browser.ts
ruyadorno: done! ✅

src/graph/src/reify/index.ts
ruyadorno: I don't think we should have this guard on reify at all, as this is very much a hot-path for installs 🤔 we'll work on removing this

lukekarrys (approved):

PR #1731, opened 2026-08-06
chore: remove @astrojs/vercel from www/docs

usrrname:
Removes the @astrojs/vercel dependency from www/docs package.json and regenerates the lock file.

## Why the lockfile was regenerated with source vlt

The initial lockfile change (commit 62411978) was regenerated with the released vlt 1.0.1, which writes a different lockfile format than the source build used by CI — it omits the `registry`/`registries` block and produces bloated peer sets (up to `~peer.133` vs the source build's `~peer.84`).

CI builds vlt from source, then reinstalls (`rm-node-nodules` + `vlt install`), which rewrote the committed lockfile — leaving the git tree dirty and failing the "Consistent Workspaces" check.

This PR regenerates `vlt-lock.json` with the source vlt build so the committed lockfile is idempotent under CI's install sequence. Verified locally: the CI sequence now produces zero diff against the committed lockfile.


usrrname:
I think we may end up needing this due to the markdown/content negotiation PR I have going on #1734

PR #1746, opened 2026-08-15
feat(cli-sdk): add audit command with `--audit-level=<severity>`

### LINKED ISSUE #1676: [FEATURE] `vlt audit` / `vlt query` could provide more diagnostic details about the malware, vulns, scripts or dependencies in question
### 📋 Description

As a developer who isn't familiar with DSS, I would like to see more diagnostic information from commands I run to further investigate potential security issues. 

I was looking into how the DSS and `vlt query` works, within my pet project pnpm workspace, badger.
It is a workspace with a packages/ folder, and 3 package children. 

Approach:
Like any lazy dev, I tried to run the commands from the root of a pnpm workspace with a `packages/` folder. 

The following are just examples recording my first impressions.

From the blog posts, ["Introducing Phased package installations"](https://www.vlt.io/blog/vlt-build), I tried running the following commands:

1. `vlt query :malware` 

The response shows a list of dependencies but I can't tell what their severity or impact is, or what kind of malware it is.

```
❯ vlt query ":malware"
badger
└─┬ @usrrname/badger-fetch@0.1.0-beta.3
  └─┬ promptfoo@0.121.15
    ├─┬ @playwright/browser-chromium@1.61.1
    │ └── playwright-core@1.61.1
    └─┬ playwright@1.61.1
      └── playwright-core@1.61.1
```

Suggested improvements:

- count of total malicious 
- identification of the malicious dependencies and versions that are most critical 
   - whether the dependency is direct, transitive, optional 
- commands to run to further diagnose what might be the most pressing
- link to doc for further help on how to mitigate

2. Provide help text for users/LLMs to recover and continue from incorrect input

When I type an incorrect flag or selector in a DSS query, it might be helpful to add a "Did you mean <correct command> " or respond with the available flags, or to include a link  to the corresponding doc page. 

For example, in trying to look for "medium" CVEs:

```
vlt query ":malware(moderate)"

Error: Failed to parse :malware selector

Full details written to: /Users/jenc/Library/Application Support/vlt/error-logs/error-6534.log

Open an issue with the full error details at:
https://github.com/vltpkg/vltpkg/issues/new
```

3. Then I try `vlt query :malware(low)` and receive the following response: 

<details> 

<summary>
Console Response 
</summary>

```
badger
├─┬ @usrrname/badger-fetch@0.1.0-beta.3
│ ├─┬ @modelcontextprotocol/sdk@1.29.0
│ │ ├── ajv@8.20.0
│ │ ├─┬ ajv-formats@3.0.1
│ │ │ └── ajv@8.20.0
│ │ ├─┬ cross-spawn@7.0.6
│ │ │ └─┬ which@2.0.2
│ │ │   └── isexe@2.0.0
│ │ ├─┬ express@5.2.1
│ │ │ ├─┬ proxy-addr@2.0.7
│ │ │ │ └── ipaddr.js@1.9.1
│ │ │ ├─┬ qs@6.15.3
│ │ │ │ └─┬ side-channel@1.1.1
│ │ │ │   ├─┬ side-channel-map@1.0.1
│ │ │ │   │ ├─┬ call-bound@1.0.4
│ │ │ │   │ │ └── get-intrinsic@1.3.0
│ │ │ │   │ └── get-intrinsic@1.3.0
│ │ │ │   └─┬ side-channel-weakmap@1.0.2
│ │ │ │     └── get-intrinsic@1.3.0
│ │ │ ├── send@1.2.1
│ │ │ └─┬ serve-static@2.2.1
│ │ │   └── send@1.2.1
│ │ ├── zod@4.4.3
│ │ └─┬ zod-to-json-schema@3.25.2
│ │   └── zod@4.4.3
│ ├─┬ @opencode-ai/plugin@1.15.13
│ │ ├─┬ @opencode-ai/sdk@1.15.13
│ │ │ └── cross-spawn@7.0.6
│ │ ├─┬ effect@4.0.0-beta.66
│ │ │ ├─┬ msgpackr@1.12.1
│ │ │ │ └─┬ msgpackr-extract@3.0.4
│ │ │ │   └── node-gyp-build-optional-packages@5.2.2
│ │ │ └── yaml@2.9.0
│ │ └── zod@4.1.8
│ ├─┬ @openrouter/devtools@0.1.0
│ │ ├─┬ @openrouter/sdk@0.3.1
│ │ │ └── zod@4.4.3
│ │ └── zod@4.1.12
│ ├─┬ @openrouter/sdk@0.12.79
│ │ └── zod@4.4.3
│ ├─┬ @types/sanitize-html@2.16.1
│ │ └── htmlparser2@10.1.0
│ ├─┬ promptfoo@0.121.15
│ │ ├─┬ @anthropic-ai/claude-agent-sdk@0.3.150
│ │ │ └── zod@4.4.3
│ │ ├─┬ @anthropic-ai/sdk@0.98.0
│ │ │ └── zod@4.4.3
│ │ ├─┬ @aws-sdk/client-bedrock-agent-runtime@3.1088.0
│ │ │ ├─┬ @aws-sdk/core@3.975.3
│ │ │ │ ├── @smithy/core@3.29.4
│ │ │ │ └─┬ @smithy/signature-v4@5.6.5
│ │ │ │   └── @smithy/core@3.29.4
│ │ │ ├─┬ @aws-sdk/credential-provider-node@3.972.69
│ │ │ │ ├─┬ @aws-sdk/credential-provider-env@3.972.59
│ │ │ │ │ └── @smithy/core@3.29.4
│ │ │ │ ├─┬ @aws-sdk/credential-provider-http@3.972.61
│ │ │ │ │ └── @smithy/core@3.29.4
│ │ │ │ ├─┬ @aws-sdk/credential-provider-ini@3.973.3
│ │ │ │ │ ├─┬ @aws-sdk/credential-provider-login@3.972.65
│ │ │ │ │ │ └── @smithy/core@3.29.4
│ │ │ │ │ └── @smithy/core@3.29.4
│ │ │ │ ├─┬ @aws-sdk/credential-provider-process@3.972.59
│ │ │ │ │ └── @smithy/core@3.29.4
│ │ │ │ ├─┬ @aws-sdk/credential-provider-web-identity@3.972.65
│ │ │ │ │ └── @smithy/core@3.29.4
│ │ │ │ ├── @smithy/core@3.29.4
│ │ │ │ └─┬ @smithy/credential-provider-imds@4.4.9
│ │ │ │   └── @smithy/core@3.29.4
│ │ │ ├── @smithy/core@3.29.4
│ │ │ └─┬ @smithy/fetch-http-handler@5.6.6
│ │ │   └── @smithy/core@3.29.4
│ │ ├─┬ @aws-sdk/client-bedrock-runtime@3.1088.0
│ │ │ ├─┬ @aws-sdk/eventstream-handler-node@3.972.28
│ │ │ │ └── @smithy/core@3.29.4
│ │ │ ├─┬ @aws-sdk/middleware-eventstream@3.972.24
│ │ │ │ └── @smithy/core@3.29.4
│ │ │ ├─┬ @aws-sdk/middleware-websocket@3.972.41
│ │ │ │ └── @smithy/core@3.29.4
│ │ │ ├─┬ @aws-sdk/token-providers@3.1088.0
│ │ │ │ └── @smithy/core@3.29.4
│ │ │ └── @smithy/core@3.29.4
│ │ ├─┬ @aws-sdk/client-s3@3.1088.0
│ │ │ ├─┬ @aws-sdk/checksums@3.1000.18
│ │ │ │ └── @smithy/core@3.29.4
│ │ │ ├─┬ @aws-sdk/middleware-sdk-s3@3.972.64
│ │ │ │ └── @smithy/core@3.29.4
│ │ │ └── @smithy/core@3.29.4
│ │ ├─┬ @aws-sdk/client-sagemaker-runtime@3.1088.0
│ │ │ └── @smithy/core@3.29.4
│ │ ├─┬ @aws-sdk/credential-provider-sso@3.973.3
│ │ │ ├─┬ @aws-sdk/nested-clients@3.997.33
│ │ │ │ └── @smithy/core@3.29.4
│ │ │ └── @smithy/core@3.29.4
│ │ ├── @azure/ai-projects@2.3.1
│ │ ├─┬ @azure/identity@4.13.1
│ │ │ └── open@10.2.0
│ │ ├─┬ @googleapis/sheets@13.0.2
│ │ │ └─┬ googleapis-common@8.0.2
│ │ │   ├── extend@3.0.2
│ │ │   └─┬ google-auth-library@10.5.0
│ │ │     └── ecdsa-sig-formatter@1.0.11
│ │ ├─┬ @ibm-cloud/watsonx-ai@1.7.15
│ │ │ └─┬ form-data@4.0.6
│ │ │   ├── asynckit@0.4.0
│ │ │   ├─┬ combined-stream@1.0.8
│ │ │   │ └── delayed-stream@1.0.0
│ │ │   ├─┬ es-set-tostringtag@2.1.0
│ │ │   │ └─┬ get-intrinsic@1.3.0
│ │ │   │   ├─┬ call-bind-apply-helpers@1.0.2
│ │ │   │   │ └── function-bind@1.1.2
│ │ │   │   └── function-bind@1.1.2
│ │ │   ├─┬ hasown@2.0.4
│ │ │   │ └── function-bind@1.1.2
│ │ │   └── mime-types@2.1.35
│ │ ├─┬ @ibm-generative-ai/node-sdk@3.2.4
│ │ │ └── yaml@2.9.0
│ │ ├─┬ @inquirer/core@11.2.1
│ │ │ └── signal-exit@4.1.0
│ │ ├─┬ @openai/agents@0.11.8
│ │ │ ├─┬ @openai/agents-core@0.11.8
│ │ │ │ └── zod@4.4.3
│ │ │ ├─┬ @openai/agents-openai@0.11.8
│ │ │ │ └── zod@4.4.3
│ │ │ ├─┬ @openai/agents-realtime@0.11.8
│ │ │ │ └── zod@4.4.3
│ │ │ └── zod@4.4.3
│ │ ├─┬ @opencode-ai/sdk@1.18.2
│ │ │ └── cross-spawn@7.0.6
│ │ ├─┬ @playwright/browser-chromium@1.61.1
│ │ │ └── playwright-core@1.61.1
│ │ ├─┬ @smithy/node-http-handler@4.9.6
│ │ │ └── @smithy/core@3.29.4
│ │ ├─┬ ai@6.0.228
│ │ │ ├─┬ @ai-sdk/gateway@3.0.151
│ │ │ │ └── zod@4.4.3
│ │ │ ├─┬ @ai-sdk/provider-utils@4.0.39
│ │ │ │ └── zod@4.4.3
│ │ │ └── zod@4.4.3
│ │ ├── ajv@8.20.0
│ │ ├─┬ gcp-metadata@8.1.2
│ │ │ └─┬ gaxios@7.1.3
│ │ │   ├── extend@3.0.2
│ │ │   └─┬ rimraf@5.0.10
│ │ │     └─┬ glob@10.5.0
│ │ │       ├─┬ foreground-child@3.3.1
│ │ │       │ ├── cross-spawn@7.0.6
│ │ │       │ └── signal-exit@4.1.0
│ │ │       └─┬ path-scurry@1.11.1
│ │ │         └── lru-cache@10.4.3
│ │ ├── glob@13.0.6
│ │ ├─┬ google-auth-library@10.9.0
│ │ │ ├── ecdsa-sig-formatter@1.0.11
│ │ │ ├─┬ gaxios@7.2.0
│ │ │ │ └── extend@3.0.2
│ │ │ └─┬ jws@4.0.1
│ │ │   └─┬ jwa@2.0.1
│ │ │     └── ecdsa-sig-formatter@1.0.11
│ │ ├─┬ ibm-cloud-sdk-core@5.5.0
│ │ │ ├── extend@3.0.2
│ │ │ └── mime-types@2.1.35
│ │ ├── json5@2.2.3
│ │ ├─┬ langfuse@3.38.20
│ │ │ └─┬ langfuse-core@3.38.20
│ │ │   └── mustache@4.2.0
│ │ ├─┬ natural@8.1.1
│ │ │ ├─┬ mongoose@9.7.4
│ │ │ │ ├─┬ mongodb@7.2.0
│ │ │ │ │ └─┬ @mongodb-js/saslprep@1.4.12
│ │ │ │ │   └── sparse-bitfield@3.0.3
│ │ │ │ └── sift@17.1.3
│ │ │ ├─┬ pg@8.22.0
│ │ │ │ └─┬ pg-types@2.2.0
│ │ │ │   └── postgres-date@1.0.7
│ │ │ └── underscore@1.13.8
│ │ ├─┬ nunjucks@3.2.4
│ │ │ ├── asap@2.0.6
│ │ │ └── commander@5.1.0
│ │ ├─┬ openai@6.47.0
│ │ │ └── zod@4.4.3
│ │ ├─┬ ora@9.4.1
│ │ │ └─┬ cli-cursor@5.0.0
│ │ │   └─┬ restore-cursor@5.1.0
│ │ │     └── signal-exit@4.1.0
│ │ ├─┬ pdf-parse@2.4.5
│ │ │ └── pdfjs-dist@5.4.296
│ │ ├─┬ pem@1.14.8
│ │ │ ├─┬ md5@2.3.0
│ │ │ │ └── crypt@0.0.2
│ │ │ └── which@2.0.2
│ │ ├─┬ playwright@1.61.1
│ │ │ └── playwright-core@1.61.1
│ │ ├─┬ playwright-extra@4.3.6
│ │ │ └── playwright@1.61.1
│ │ ├─┬ posthog-node@5.24.17
│ │ │ └─┬ @posthog/core@1.23.1
│ │ │   └── cross-spawn@7.0.6
│ │ ├─┬ proxy-agent@8.0.2
│ │ │ └─┬ pac-proxy-agent@9.1.0
│ │ │   ├─┬ pac-resolver@9.0.1
│ │ │   │ ├─┬ degenerator@7.0.1
│ │ │   │ │ ├─┬ escodegen@2.1.0
│ │ │   │ │ │ └── esprima@4.0.1
│ │ │   │ │ ├── esprima@4.0.1
│ │ │   │ │ └── quickjs-wasi@2.2.0
│ │ │   │ └── quickjs-wasi@2.2.0
│ │ │   └── quickjs-wasi@2.2.0
│ │ ├── python-shell@5.0.0
│ │ ├─┬ read-excel-file@9.3.1
│ │ │ └─┬ unzipper-esm@0.13.2
│ │ │   └── graceful-fs@4.2.11
│ │ ├── rxjs@7.8.2
│ │ ├─┬ socket.io@4.8.3
│ │ │ └─┬ accepts@1.3.8
│ │ │   └── mime-types@2.1.35
│ │ ├─┬ tsx@4.23.1
│ │ │ └── esbuild@0.28.1
│ │ ├── undici@7.28.0
│ │ ├─┬ winston@3.19.0
│ │ │ └── stack-trace@0.0.10
│ │ └── zod@4.4.3
│ ├─┬ sanitize-html@2.17.5
│ │ ├── htmlparser2@10.1.0
│ │ └─┬ postcss@8.5.19
│ │   └── source-map-js@1.2.1
│ ├─┬ vitest@3.2.6
│ │ ├─┬ vite@7.3.6
│ │ │ ├── esbuild@0.28.1
│ │ │ └── rollup@4.62.2
│ │ └─┬ vite-node@3.2.4
│ │   └── es-module-lexer@1.7.0
│ └── zod@4.4.3
├─┬ @usrrname/badger-response@0.1.0-beta.2
│ └── zod@4.4.3
├─┬ bumpp@11.1.0
│ ├─┬ unconfig@7.5.0
│ │ └── jiti@2.7.0
│ └── yaml@2.9.0
└─┬ git-cliff@2.13.1
  └─┬ execa@9.6.1
    ├── cross-spawn@7.0.6
    └── signal-exit@4.1.0
```

</details>

The full list is rather long and intimidating though it lists all the low severity malware inside my whole workspace. 

Suggested improvement:
- Display total number of low severity issues found in a workspace 
- Display how many low sev issues are direct dependencies within the workspace
- Display how many are indirect dependencies within the workspace
- Provide a 1 sentence explanation to user what vlt considers malware and a link to read more .

ie. 

```
30 low severity malware in @usrrname/badger
15 low severity malware-containing dependencies in `@usrrname/badger/packages/badger-fetch`
...
```

4. When I run `vlt query :scripts` in search of lifecycle scripts within a codebase (assuming I'm concerned to debug any potential ones that are running out of control), I would like to know the type of lifecycle script that is in the particular dependency's manifest.

```
❯ vlt query ":scripts"
badger
└─┬ @usrrname/badger-fetch@0.1.0-beta.3
  ├─┬ @opencode-ai/plugin@1.15.13
  │ └─┬ effect@4.0.0-beta.66
  │   └─┬ msgpackr@1.12.1
  │     └── msgpackr-extract@3.0.4
  ├── @openrouter/sdk@0.12.79
  ├─┬ promptfoo@0.121.15
  │ ├── @playwright/browser-chromium@1.61.1
  │ ├── @swc/core@1.15.43
  │ ├─┬ playwright@1.61.1
  │ │ └── fsevents@2.3.2
  │ ├── sharp@0.34.5
  │ └─┬ tsx@4.23.1
  │   ├── esbuild@0.28.1
  │   └── fsevents@2.3.3
  └─┬ vitest@3.2.6
    └─┬ vite@7.3.6
      ├── esbuild@0.28.1
      ├── fsevents@2.3.3
      └─┬ rollup@4.62.2
        └── fsevents@2.3.3
```

I imagine developers might like to see:
- a count of lifecycle scripts within direct dependencies (and what kind)
- counts of lifecycle scripts in indirect dependencies 
- which ones may be and any follow up commands I can run to get a finer grained search 
- what commands I could run to further investigate 

### 🔍 Research & Prior Art

This is probably totally trodden territory for you all, but nonetheless, gotta complete the ticket!

1.  clig.dev / https://github.com/cli-guidelines/cli-guidelines

> Human-readable output is paramount. Humans come first, machines second. The most simple and straightforward heuristic for whether a particular output stream (stdout or stderr) is being read by a human is whether or not it’s a TTY. Whatever language you’re using, it will have a utility or library for doing this (e.g. [Python](https://stackoverflow.com/questions/858623/how-to-recognize-whether-a-script-is-running-on-a-tty), [Node](https://nodejs.org/api/process.html#process_a_note_on_process_i_o), [Go](https://github.com/mattn/go-isatty)).
...

> Make it easy to see the current state of the system. If your program does a lot of complex state changes and it is not immediately visible in the filesystem, make sure you make this easy to view.

Image

> Display output on success, but keep it brief. Traditionally, when nothing is wrong, UNIX commands display no output to the user. This makes sense when they’re being used in scripts, but can make commands appear to be hanging or broken when used by humans. For example, cp will not print anything, even if it takes a long time ...It’s rare that printing nothing at all is the best default behavior, but it’s usually best to err on the side of less.

> For instances where you do want no output (for example, when used in shell scripts), to avoid clumsy redirection of stderr to /dev/null, you can provide a -q option to suppress all non-essential output.

2. I do not know that this is great prior art, but it seems in our territory: 
https://github.com/npm/metavuln-calculator

> Dependent Version Sampling
> Typically, dependency ranges don't change very frequently, and the most recent version published on a given release line is most likely to contain the fix for a given vulnerability.

> So, we see things like this:
```
3.0.4 - not vulnerable
3.0.3 - vulnerable
3.0.2 - vulnerable
3.0.1 - vulnerable
3.0.0 - vulnerable
2.3.107 - not vulnerable
2.3.106 - not vulnerable
2.3.105 - vulnerable
... 523 more vulnerable versions ...
2.0.0 - vulnerable
1.1.102 - not vulnerable
1.1.101 - vulnerable
... 387 more vulnerable versions ...
0.0.0 - vulnerable
```
3. npm audit examples

https://docs.npmjs.com/auditing-package-dependencies-for-security-vulnerabilities#no-security-vulnerabilities-found 

Image

4. OSV scanner CLI with interactive mode 

Image

Proof of concept:

#1700

usrrname:
## Summary

Implements `vlt audit` command for security scanning with npm/pnpm-compatible output.

Screenshot 2026-08-28 at 12 48 01 PM

## Features

- **`vlt audit`** — Scan installed dependencies for security issues (vulnerabilities, malware, typosquats)
- **`--audit-level`** — Filter by severity: `low` (default), `medium`, `high`, `critical`
- **Multiple output formats** — Human-readable (default), JSON (`--view=json`), counts (`--view=count`), silent
- **CVE links** — Click-through links to NVD vulnerability details (in supported terminals)
- **Security summary in `vlt query`** — footer shows security findings when querying with security selectors
- `:script` is not considered part of security selector queries as our internal codebase nor socket provides the judgment around lifecycle hooks or code execution

### Pipeline

`commands/audit.ts` builds a DSS query from `--audit-level` (`buildAuditQuery`),
runs it, and passes the matched nodes to `aggregateBySeverity`, which buckets
packages by severity. `filterAuditResult` then drops buckets below the requested
level and recomputes the direct/transitive split from what survives. Views
render from the resulting `AuditResult`.

`vlt query`'s security footer calls the **same aggregator** — but with no
security archive. 

Screenshot 2026-08-17 at 2 26 16 PM

- N packages with security issues found — counts nodes, one increment per node that yielded at least one alert. Not alerts, not edges. A package with six CVEs counts once to prevent overcounting.

  - critical (2), high (1), … — each node is filed in exactly one bucket(its worst severity), so these are mutually exclusive and sum to the total.
  - N direct dependencies, M transitive — also per package, also summing to the total. "Direct" means the node is itself an importer, or appears in some importer's edgesOut.
 
### Severity model

Severity is taken from each alert's own `severity` field, not inferred from its
type name. The Socket feed grades CVEs in the type name *as well* (`mildCVE`,
`mediumCVE`, `highCVE`, `criticalCVE`), and type-name mapping missed the grades
nobody had enumerated. `typeImpliedLevel` remains as a fallback when an alert
arrives without a usable severity, so behaviour is preserved for those.

- The wire spelling is **`middle`**, never `medium`; every consumer normalizes.
  `AlertSeverity` deliberately excludes `medium` so this can't be forgotten.
- An unrecognized severity is **not** allowed to fall through to `low` — malware
  is treated as critical, anything else as low, and `warn()` reports it. Without
  this, a critical malware alert whose severity the feed spelled unexpectedly
  would be filed `low`, dropped by `--audit-level=high`, and exit `0`.
- Synthetic alerts emit `middle` via `toWireSeverity`, so a synthetic
  `PackageAlert` never carries a value the endpoint can't produce.

### Alert filtering and classification

- Alerts the feed marks **`action: "ignore"`** are not surfaced — no row, no
  count. This is the large majority of alerts on a normal tree (capability
  signals like `envVars`/`networkAccess`), and reporting them buried the
  minority that matter. Filtered at the single point alerts enter
  (`getPackageAlerts`) so bucketing, counts, and rows can't disagree.
- Malware and typosquats are matched on `alert.type`, not `category` — both sit
  under `supplyChainRisk` alongside every capability finding, so category alone
  can't separate them.
- Counts are **per package**, matching the severity line beneath them; a package
  counts at most once per category.
- Scan coverage is measured over the **whole graph**, not the audit's results
  (every result had feed data by definition), and survives filtering.

### Remediation

- `fix` is present if a fix exists, so its absence is authoritative rather than
  inferred. `fix.description` is **not** rendered since it's too long to be printed
- Suggested versions are exact, never `>=`, and anything interpolated into a
  copy-pasteable command is validated (semver for the version, npm's grammar for
  the name).
- A fix command is offered only for **direct** dependencies. Transitive findings get
  reachability computed locally from the ranges their dependents declare
  (`@vltpkg/semver`) — no registry round-trip.
- CISA KEV findings are badged, counted in the summary, and **exempt from
  `--audit-level` filtering**. Their severity is not rewritten, which would
  contradict the feed's own grading.

### Terminal safety

Advisory text, package names, and versions are third-party input written to a
terminal, where control bytes are instructions. New `src/cli-sdk/src/safe-text.ts`
provides `safeText`/`safeUrl`:

- strips C0/C1 controls, DEL, and bidi overrides, and length-caps.
  `node:util`'s `stripVTControlCharacters` is **not** sufficient — it leaves a
  bare `ESC`, `BEL`, `\r` and `\n`, which is enough to terminate an OSC 8
  hyperlink early and emit arbitrary sequences (clipboard writes via OSC 52, a
  forged "0 packages with security issues" line).
- hyperlink targets are restricted to `http`/`https`, since a terminal
  supporting OSC 8 hands the target to the OS opener.
- a feed-supplied URL is labelled with its **host**; only a URL we construct
  ourselves is labelled with an advisory id, so a link can't misrepresent where
  it goes.
- zero-allocation fast path: it tests before rewriting and returns the original
  when already clean.

### New in `@vltpkg/graph`

`src/graph/src/dependency-paths.ts` — `getDependencyPaths`, `getDependents`,
`countDependencyPaths`. Answers "why is this installed" by walking `edgesIn`.

Two **independent** bounds, because a paths-found cap alone doesn't bound the
work: cycles terminate via a per-branch ancestor set, but a strongly connected
cluster with no route to an importer dead-ends without recording a path, so the
cap never fires while every simple path inside is enumerated. `maxVisits` caps
that. The walk mutates one shared trail/ancestor set and restores on the way back
up — copying both per edge made a single branch O(depth²). Two edges sharing a
parent can't yield the same route twice.

### Output

Alert types render in `npm audit`/`pnpm audit` language rather than feed
vocabulary (`urlStrings` → `embedded URL`, `gptAnomaly` →
`code anomaly (AI-flagged)`), via a lookup table that falls back to the raw type
so a newly added upstream alert still reports. Each finding's advisory ids sit on
its own line — a flat per-package list loses which CVE belongs to which finding.
Identical findings collapse with an `xN` count. The summary renders after the
findings.

### Other fixes in this PR

- `--audit-level=moderate` now genuinely resolves to `medium`. jackspeak has no
  `normalize` hook (only `validate`), so the value previously reached the command
  as typed, matched no branch, and fell through to `low` — reporting every
  low-severity finding.
- `:malware` appears unqualified at every level (it takes no severity
  parameter), or malware-only findings are dropped from the query entirely.
- `:severity()` in `@vltpkg/query` grades from the alert's severity field,
  scoped to the Vulnerability category. Its kind→type map paired one type per
  level, so `mediumCVE`/`highCVE` matched nothing.
- `vlt view` rendered medium-severity alerts as `[MIDDLE]` — it handled
  `'medium'` but the feed sends `'middle'`.

## Breaking changes

All four affected packages are published at 1.0.2 (none private).

- **`@vltpkg/query`** — `Insights.malware` is now `LeveledInsights` rather than
  `boolean`. This changes `vlt query --view=json` output shape, and is a *silent*
  break for a consumer doing `if (insights.malware)`, which now always sees a
  truthy object.
- **`@vltpkg/security-archive`** — `AlertCategory` now holds exactly the six
  categories the endpoint documents; `malware`, `severity` and `squat` are alert
  *types*, not categories, and `info` is now `other`. `PackageAlert.severity` and
  `.category` are optional, and `AlertSeverity` excludes `medium`.
- **`:severity()`** matches more than before (`mediumCVE`/`highCVE` previously
  matched nothing), so existing queries return a superset rather than losing
  results.

Types were verified field-by-field against Socket's documented schema:
<https://docs.socket.dev/reference/batchpackagefetch>. Previously-discarded
documented fields are now declared: `action`, `fix`, `actionSource`,
`actionPolicyIndex`, `patch`, `reachability`, `subType`, `file`/`start`/`end`.

## Known limitations

- **Findings describe mechanism, not impact.** There is no RCE category; it
  arrives either inside a CVE's title/CWEs or as capability signals that are
  preconditions for it. Audit reports the capability and infers nothing.
- **Discovery is narrower than reporting.** A package is found via malware,
  typosquat, or vulnerability; other findings only show up once it's been found
  another way. Widening it means adding bounded binary selectors and extending
  the category counters — `:license` must stay out, as it matches nearly the
  whole graph only for the ignore filter to discard it.
- `props.epss` and `reachability` are declared but unused.

## Testing

Automated tests were included for:
- audit command
- query
- `@vltpkg/graph` dependency paths
- `:severity` selector grading

```
vlr test -w src/cli-sdk
vlr test -w src/query
vlr test -w src/graph
vlr test -w src/security-archive
```

Note that `vlr test` also runs each workspace's `posttest: tsc --noEmit` and the
100 % coverage gate, so **judge it by its exit code** — both of those report
below tap's summary line.

## Testing locally with this branch checked out

From the worktree root, use the repo's own bin shim — not a bare vlt:
```
    cd /Users/jenc/code/vlt/vltpkg/.claude/worktrees/jc-display-number-of-malware

    ./scripts/bins/vlt audit
    ./scripts/bins/vlt audit --audit-level=high     # critical + high only
    ./scripts/bins/vlt audit --audit-level=moderate # alias for medium
    ./scripts/bins/vlt audit --view=json
    ./scripts/bins/vlt audit --view=count

    ./scripts/bins/vlt query ':malware'
    ./scripts/bins/vlt query ':squat'
    ./scripts/bins/vlt query ':severity(critical)'
```

 Plain vlt audit runs the published 1.0.2 or your most recent vlt upgrade, which has none of this branch's work. `./scripts/bins/vlt` runs the branch source. If output looks unchanged after an edit, that's the  first thing to check.

 audit sets `needsRegistry = true` and hits Socket's batch endpoint, so it needs network and a logged-in session to return real findings. Against a tree with nothing flagged you'll get the "no issues" path plus the scan-coverage line — fine for smoke-testing formatting, useless for testing severity buckets.

Useful things to eyeball: the summary printing after the findings, moderate genuinely filtering to medium (the bug was it silently fell through to low), and xN collapsing on a package like sharp or node-gyp.

    Running the tests

    `vlr test -w src/cli-sdk`

    Judge it by echo $?, not by the assertion tally. It exits non-zero for the posttest tsc check and for the 100% coverage gate, and both print below a cheerful "TEST COMPLETE" — a green-looking run can still be a failure.

    For the coverage table on demand:
```
    cd src/cli-sdk
    ./node_modules/.bin/tap --coverage-report=text --show-full-coverage
```

    Note tap lives in `src/cli-sdk/node_modules/.bin/`, not the root `node_modules`. The uncovered-lines column gets truncated at terminal width; `./node_modules/.bin/tap report --coverage-report=lcovonly` re-reports from the last run's data  without re-running, writing `.tap/report/lcov.info` with exact per-branch hits.

---

# `vlt audit`

Check installed dependencies for security issues — malware, vulnerable
packages, and typosquats.

```bash
vlt audit
vlt audit --audit-level=high
vlt audit --view=json
```

| option          | values                                              | default |
| --------------- | --------------------------------------------------- | ------- |
| `--audit-level` | `low`, `medium` (or `moderate`), `high`, `critical` | `low`   |
| `--view`        | `human`, `json`, `count`                            | `human` |

Exits `1` when any finding is reported, `0` when none is, matching
npm/pnpm so it can gate CI directly.

## Output

```
high  nanoid@3.3.16
        vulnerability: high  CVE-2026-67213  GHSA-2v37-7h3g-55p8  CVSS 8.2
          nanoid: custom generators can loop indefinitely when size is zero
          affects < 3.3.18 -- patched in 3.3.18
        via @vltpkg/docs > @astrojs/tailwind > postcss > nanoid

3 packages with security issues        (a)
1 malware, 2 vulnerable                (b)
1 critical, 2 high                     (c)
1 actively exploited                   (d)

1 direct dependency, 2 transitive      (e)
238 of 240 installed versions scanned  (f)
```

A header line per package with its worst severity, then one line per
distinct finding, then where it sits in the graph. Every summary line
counts **packages** except (f), which counts installed versions.

- **(a)** packages with at least one finding, post-filtering
- **(b)** packages per category; a package with both a CVE and a
  malware alert adds to each, so (b) can exceed (a)
- **(c)** packages at their worst severity — mutually exclusive, so
  (c) always sums to (a)
- **(d)** packages with an actively exploited finding, a subset of (a)
- **(e)** direct vs transitive, summing to (a)
- **(f)** scan coverage over the whole graph

## Requirements

- Report malware, vulnerabilities, and typosquats across the installed
  graph, with a severity floor for CI use.
- Every finding must carry an actionable next step, or say plainly
  that there isn't one.
- Never let a severe finding disappear because of a filter, a spelling
  mismatch, or a default.
- Never claim more confidence than the data supports — including about
  how much of the tree was actually scanned.
- Terminal output is untrusted third-party text and must be safe to
  render.

### Severity

**Severity comes from each alert's own `severity` field**, and a
package is filed at its worst. The wire scale spells the middle level
`middle`; left untranslated it matched no rank, so a package whose
worst finding was `middle` was filed under `low`.

**An unrecognized severity is not defaulted to `low`.** Malware is
treated as `critical`, everything else `low`, and a warning is emitted
either way. The default would have been silent data loss: a feed
spelling a malware severity `moderate` would vanish under
`--audit-level=high` and the command would exit `0`.

**`moderate` is an npm-compatible alias for `medium`, normalized in
`commands/audit.ts`, not `definition.ts`.** jackspeak offers a
`validate` hook but no `normalize`, so the value arrives as typed;
`definition.ts` only accepts it. Left unresolved it fell through to
the `low` default, reporting every low-severity finding under
`--audit-level=moderate`.

### Filtering

**Alerts the feed marks `action: "ignore"` are never surfaced** — no
row, no count. They are the large majority (capability signals like
`envVars` on ordinary packages) and bury real findings. Filtered in
`getPackageAlerts`, the single point alerts enter, so bucketing,
counts, and rendered rows cannot disagree about which alerts exist. A
package whose every alert is ignored drops out entirely.

**CISA KEV findings are exempt from `--audit-level`.** They keep the
severity the feed assigned — we don't second-guess it — but are
reported, badged, and counted below the threshold, so a gate on
`--audit-level=critical` can't silently discard something being
exploited right now. It is also called out in the summary, since a
badge on one detail line is easily scrolled past.

**Discovery is narrower than reporting.** The query matches only
malware, vulnerability, and squat selectors; every other alert on a
matched package is then reported in full. `:scripts` was deliberately
excluded — nothing consumes lifecycle-script findings, so matching
them did work for no visible output. Not widened to `:obfuscated`,
`:unmaintained`, or `:deprecated`: a package whose only issue is one
of those does not appear at all.

### Counting

**Counts are per package, never per alert.** They render directly
above a line that counts packages, and adjacent count lines read as
the same unit. Counting alerts also multiplied single findings — one
CVE sets both `insights.severity` and `insights.vuln`, so the
synthetic path emits two alerts for it.

**The summary prints after the findings**, since the counts are the
conclusion you read once the detail has scrolled past.

**Scan coverage is measured over the whole graph, not the results.**
Every result matched a security selector and so by definition had feed
data; counting results would claim near-total coverage whatever the
truth. It survives `--audit-level` filtering, because it describes the
graph rather than the findings. Always printed — "no issues found"
means much less if a tenth of the tree was never looked at.

### Presentation

**Path counts, not path lists.** One route prints inline as
`via a > b > c`; beyond that, `reached by N paths`. Listing forty
routes buries the finding, and showing one arbitrary route understates
it, since acting on that route alone leaves the version installed via
the rest. A truncated count renders `N+`, never a bare `N`.

**Advisory output is title, CVSS score, advisory link, and
affected/patched versions — not the description**, which runs to
paragraphs and belongs on the linked page. The feed's
`fix.description` is also dropped: it reads "run `npx socket fix`",
pointing at another package manager.

**Only direct dependencies get a copy-pasteable install command**,
pinned to the exact patched version. A transitive package's version is
chosen by its dependents, and we have not verified a release of the
dependent exists whose range admits the patch, so `reachable:` /
`blocked:` reports what the local graph does establish and leaves the
call to the reader. A suggested version must parse as semver before it
reaches a command line.

**Findings describe mechanism, not impact.** There is no "remote code
execution" category. `shell access` + `network access` +
`install script` is equally the shape of a malicious postinstall and
of a legitimate native build, so audit reports the capability and
infers nothing about its use.

**Findings group and collapse by alert `type`, not `category`.**
Category is the coarse grouping — every supply-chain finding is
`supplyChainRisk` — so grouping by it rendered distinct alerts as
identical lines that collapsed into a meaningless count. Malware is
likewise matched on `type`: filed under a supply-chain category, it
counted nowhere, so detail rows read "malware: critical" while the
headline said zero.

**Alert types are relabeled for humans.** `urlStrings` and
`gptDidYouMean` are API vocabulary. Unknown types fall back to the raw
type, so a newly added upstream alert still reports rather than
vanishing.

**Links are gated on colors, not terminal detection.** With colors
off, the URL is printed beside the id, since a hidden hyperlink is
useless in piped output.

## Implementation

[`src/audit-helpers.ts`](./src/audit-helpers.ts),
[`src/commands/audit.ts`](./src/commands/audit.ts), and the
`audit-level` option in
[`src/config/definition.ts`](./src/config/definition.ts).

`vlt query`'s security-summary footer shares the same aggregator and
calls it _without_ a security archive, so the synthetic-alert fallback
in `getPackageAlerts` is load-bearing, not dead code.

Advisory text, package names, and versions are third-party input
rendered to a terminal, where control bytes are instructions rather
than characters. All of it passes through `safeText`/`safeUrl` in
[`src/safe-text.ts`](./src/safe-text.ts); `node:util`'s
`stripVTControlCharacters` is not sufficient alone, as it leaves a
bare `ESC`, `BEL`, `\r`, and `\n` intact.


vltbaudbot (commented):
## Code Review: `vlt audit` & Related Changes

Solid addition — the audit command is well-structured and the test coverage is thorough. Main feedback is around reusing existing codebase utilities and a few correctness issues.

### Reuse Existing Types & Utils

1. **Type guards in `audit-helpers.ts` (lines 131–230)** — `LeveledInsights`, `SquatInsights`, and `Insights` are already exported from `@vltpkg/query` (`src/query/src/types.ts`). Import and use them rather than hand-rolling structural type guards.

2. **Severity rank map (lines 116–121)** — duplicates `kindLevelMap` from `src/query/src/pseudo/severity.ts`. Export and reuse it to avoid drift.

3. **`styleText` wrapper (lines 8–10)** — `src/cli-sdk/src/output.ts` already has this exact pattern with `validateStream: false`. Use `styleTextStdout`/`styleTextStderr` from there.

4. **`isNodeWithEdgesOut` (lines 152–167)** — after `isNode(o)` passes (from `@vltpkg/graph`), `edgesOut: Map` is guaranteed. The extra structural check is dead code.

### Correctness Issues

5. **Hyperlink escape sequences (lines 654–665)** — Missing `\u001b` (ESC) prefix before `]8;;`. Compare with `src/url-open/src/index.ts` line 32 which correctly uses `\u001b]8;;`. The current implementation won't render clickable links in terminals.

6. **`_warn` parameter (line 391)** — `aggregateBySeverity` accepts a `_warn` callback but never invokes it. Either implement the warning calls or remove the parameter and update JSDoc. The test assertions around line 350 may be testing a codepath that no longer exists.

7. **`audit-level` normalization done in 3 places** — `config/definition.ts` validate, `config/definition.ts` normalize, AND `audit.ts` line 84 all do `moderate → medium`. Do it once in `normalize` and trust it downstream.

8. **`isVulnerabilityAlert` returns `boolean`, not a type guard** — In `security-archive/src/types.ts` (lines 117–124), should be `(alert: PackageAlert): alert is VulnerabilityPackageAlert` for proper TypeScript narrowing. Currently downstream code needs unsafe property access.

### Architecture

9. **`commands/query.ts` — direct `stdout()` calls break command/view separation** — The query command now calls `stdout()` directly for "No packages found" messages (three locations). Other commands return empty results and let the view handle rendering. Keep the pattern consistent.

10. **`needsRegistry = true` on query command (line 37)** — `vlt query` previously worked without a registry config. This will now fail with ECONFIG for local-only queries like `vlt query '*'`. Consider gating this more narrowly or making `SecurityArchive.start()` degrade gracefully.

11. **`constructSyntheticAlerts` (lines 234–295)** — This reverse-engineers `PackageAlert` objects from insights, but insights are themselves derived from alerts in `Query.#getQueryResponseNodes()`. The round-trip loses data. Since `securityArchive` is always available (started unconditionally in `audit.ts`), is this fallback reachable? If not, consider removing or documenting when it would be.

### Minor

12. **Box-drawing mixes ASCII `-` with Unicode `─`** (lines 824–826) — Use `'─'.repeat(...)` (U+2500) consistently with the `┌┬├┼└┴` characters.

13. **`categoryCounts` (lines 590–610)** — mixes `alert.type` and `alert.category` for classification. Pick one field consistently to avoid potential double-counting.

14. **Test file `test/commands/audit.ts`** — The variable assignment pattern around lines 458–530 uses a confusing parenthesization/comma-expression style. The cleaner pattern on line 829 (`node.insights = {...}`) is more readable and less error-prone.

### Overall

The audit functionality is a great addition. The main themes: (1) reuse existing types/utils from `@vltpkg/query` and `security-archive`, (2) fix the broken hyperlink escapes, (3) maintain command/view separation in the query changes.

src/cli-sdk/src/audit-helpers.ts
```diff
@@ -0,0 +1,910 @@
...
+import { isNode } from '@vltpkg/graph'
+import type { Node } from '@vltpkg/graph'
+import type { Insights } from '@vltpkg/query'
+import type { PackageAlert } from '@vltpkg/security-archive'
+import { isVulnerabilityAlert } from '@vltpkg/security-archive'
+
+const styleText = (
+  format: Parameters<typeof utilStyleText>[0],
+  s: string,
+) => utilStyleText(format, s, { validateStream: false })
```
vltbaudbot, line 26: This duplicates the `styleText` wrapper already in `output.ts`. Consider reusing it:

```suggestion
import { styleTextStdout as styleText } from './output.ts'
```

Then remove the `import { styleText as utilStyleText } from 'node:util'` on line 1.

src/cli-sdk/src/audit-helpers.ts
```diff
@@ -0,0 +1,910 @@
...
+ * gracefully when `enabled` is backed by real TTY detection --
+ * otherwise fall back to printing the URL alongside the text, since
+ * a hidden link is useless in piped/non-interactive output.
+ */
+const hyperlink = (
+  text: string,
+  url: string,
+  enabled?: boolean,
+): string =>
+  enabled ? `]8;;${url}\\${text}]8;;\\` : `${text} (${url})`
```
vltbaudbot: Missing `\u001b` (ESC) prefix before `]8;;` — these OSC 8 sequences won't render as clickable links in terminals. Compare with `src/url-open/src/index.ts` line 32.

```suggestion
  enabled ? `\u001b]8;;${url}\u001b\\${text}\u001b]8;;\u001b\\` : `${text} (${url})`
```

src/security-archive/src/types.ts
```diff
@@ -45,26 +45,87 @@ export const asSecurityArchiveLike = (
...
+  category: AlertCategory
   props?: PackageAlertProps
 }
 
+/**
+ * Type guard to check if an alert is a vulnerability/CVE alert
+ */
+export const isVulnerabilityAlert = (
+  alert: PackageAlert,
+): boolean => {
```
vltbaudbot: This returns `boolean` but is used as a type guard downstream (e.g. in `audit-helpers.ts` `.find(isVulnerabilityAlert)` then accessing `.props?.title`). Make it a proper TypeScript type guard so the compiler can narrow the type:

```suggestion
export const isVulnerabilityAlert = (
  alert: PackageAlert,
): alert is PackageAlert & { type: VulnerabilityAlertType } => {
```

vltbaudbot (commented):
Inline suggestions for the items flagged in the review above. Click 'Apply suggestion' to commit each fix directly.

src/cli-sdk/src/commands/audit.ts
```diff
@@ -0,0 +1,135 @@
...
+  })
+
+  const securityArchive = await SecurityArchive.start({
+    nodes: [...graph.nodes.values()],
+  })
+
+  const auditLevelRaw = conf.get('audit-level')
+  // Normalize 'moderate' to 'medium' for backward compatibility
+  const normalized =
+    auditLevelRaw === 'moderate' ? 'medium' : auditLevelRaw
```
vltbaudbot: This is the third place doing `moderate → medium` normalization (also in `config/definition.ts` validate and normalize). Since `conf.get('audit-level')` already returns the normalized value from the config layer, this is redundant:

```suggestion
  const auditLevel: SeverityLevel =
    (auditLevelRaw === 'critical' || auditLevelRaw === 'high' || auditLevelRaw === 'medium' || auditLevelRaw === 'low') ? auditLevelRaw : 'low'
```

src/cli-sdk/src/audit-helpers.ts
```diff
@@ -0,0 +1,910 @@
...
+})
+
+// lower number = more severe, matching kindLevelMap in
+// src/query/src/pseudo/severity.ts
+const severityRank: Record<SeverityLevel, number> = {
+  critical: 0,
+  high: 1,
+  medium: 2,
+  low: 3,
+}
```
vltbaudbot, line 173: This duplicates `kindLevelMap` from `src/query/src/pseudo/severity.ts` (and the comment even acknowledges it). Consider importing and reusing to avoid the maps drifting out of sync:

```suggestion
// Re-use the canonical severity ranking from @vltpkg/query
// rather than maintaining a duplicate map
import { kindLevelMap } from '@vltpkg/query'
const severityRank: Record<SeverityLevel, number> = kindLevelMap
```

(Verify `kindLevelMap` is exported from the package — if not, add the export rather than duplicating.)

vltbaudbot (commented):
## Re-review: `vlt audit` & Related Changes (PR #1746)

This is a re-review following the updates. The PR has grown from ~3.5k to ~7.6k additions. The author has addressed the majority of the previous findings and the code is substantially improved. CI is fully green.

---

### Previous Findings — Status

| # | Finding | Status |
|---|---------|--------|
| 1 | Type guards in `audit-helpers.ts` re-derive types from `@vltpkg/query` | ✅ **Fixed** — now imports `Insights`, `LeveledInsights`, `SquatInsights` from `@vltpkg/query` and `PackageAlert`, `AlertSeverity`, `isVulnerabilityAlert` from `@vltpkg/security-archive` |
| 2 | Severity rank map duplicates `kindLevelMap` from `severity.ts` | ⚠️ **Partially addressed** — see below |
| 3 | `styleText` wrapper duplicates `output.ts` | ⚠️ **Not addressed** — see below |
| 4 | `isNodeWithEdgesOut` redundant after `isNode()` | 🔄 **Intentionally kept** — the function has a different signature (accepts `unknown`) and guards `.edgesOut` access on nodes without that property |
| 5 | Hyperlink escape sequences broken (missing ESC prefix) | ✅ **Fixed** — raw ESC bytes now present in the template literal (verified via `cat -v`) |
| 6 | `_warn` callback accepted but never invoked | ✅ **Fixed** — now invoked via `isVerbose(conf.values.loglevel)` gating in `commands/audit.ts` |
| 7 | `audit-level` moderate→medium normalization done in 3 places | ✅ **Fixed** — now in exactly 2 places: `definition.ts` validate and `commands/audit.ts` command, which is the correct separation (jackspeak has no normalize hook) |
| 8 | `isVulnerabilityAlert` not a proper type guard | ✅ **Fixed** — now properly narrows to `PackageAlert & { type: VulnerabilityAlertType }` |
| 9 | Query command `stdout()` calls break command/view separation | ✅ **Fixed** — security summary is now rendered inside the `human` view function, not via `stdout()` |
| 10 | `needsRegistry = true` on query may break local-only queries | 🔄 **N/A** — `needsRegistry` is on `audit.ts` only (which always needs registry access); query retains its existing behavior |
| 11 | `constructSyntheticAlerts` round-trips alerts→insights→alerts | ⚠️ **Intentionally kept** — documented in AUDIT.md as load-bearing for `vlt query`'s security-summary footer which runs without a security archive |
| 12 | Box-drawing mixes ASCII `-` with Unicode `─` | ✅ **Fixed** — no more box-drawing at all; output uses plain indentation |
| 13 | `categoryCounts` mixes `alert.type` and `alert.category` | ✅ **Fixed** — now correctly classifies: malware/squat by `type` (since both sit under `supplyChainRisk` category), vulnerability by category + `isVulnerabilityAlert` fallback |
| 14 | Test quality concerns in `test/commands/audit.ts` | ✅ **Significantly improved** — added `test/audit-behaviour.ts` with real DSS engine integration tests, and the command tests now use `mockAuditWithNodes` pattern |

---

### Remaining Issues (from previous findings)

#### 1. `styleText` wrapper still duplicates `output.ts` (minor)

`audit-helpers.ts` line 23–26 defines:
```ts
const styleText = (format, s) => utilStyleText(format, s, { validateStream: false })
```

`output.ts` line 59–60 defines the identical function:
```ts
const styleText: StyleTextFn = (f, s) => utilStyleText(f, s, { validateStream: false })
```

The `output.ts` version isn't exported (it's behind `styleTextStdout`/`styleTextStderr`), so the duplication is understandable as-is. This is a nit — if `output.ts` were to export a shared `styleTextRaw` or similar, both could use it, but it's not blocking.

#### 2. Alert-type sets defined 3–4 times across the codebase

This is the main consolidation opportunity remaining:

- **`cveAlertTypes`** is defined in `src/query/src/index.ts:127`, `src/query/src/pseudo/severity.ts:57` — identical 6-element sets
- **`malwareAlertTypes`** is defined in `src/query/src/index.ts:47`, `src/cli-sdk/src/audit-helpers.ts:200` — identical 2-element sets
- **`typeImpliedLevel`** is defined as a `Record` in `src/query/src/index.ts:59` and a `Map` in `src/query/src/pseudo/severity.ts:72` — same data, different structures
- **`vulnerabilityAlertTypes`** in `src/security-archive/src/types.ts:243` is a superset of `cveAlertTypes` (adds `gptSecurity`, `gptAnomaly`)
- **`vulnAlertTypes`** in `src/query/src/index.ts:140` is the same superset

These should ideally live in one place. `@vltpkg/security-archive` already exports `isVulnerabilityAlert` and `VulnerabilityAlertType` — the sets could live there too and be imported by `@vltpkg/query` and `@vltpkg/cli-sdk`. Not blocking, but the 4-way duplication will drift.

---

### New Observations

#### ✅ Strong positives in the update

1. **`AUDIT.md` design document** — Excellent addition. Documents every design decision (severity bucketing, filter policy, counting semantics, presentation choices) with clear rationale. This is the kind of artifact that prevents future regressions.

2. **`safe-text.ts` / `safeUrl`** — Proper terminal injection defense. Strips C0/C1 controls, bidirectional overrides, and limits output length. URL filtering to http(s) only. This was not in the original and addresses a real security concern.

3. **`dependency-paths.ts` in `@vltpkg/graph`** — Clean extraction of path-finding logic into the graph package where it belongs. Proper cycle handling, bounded traversal with `maxVisits`, and the `getDependents` helper provides range data for reachability analysis. Well-tested with 465 lines of tests.

4. **Feed `middle`→`medium` normalization** — Consistently handled at every layer: `severity.ts` (`alertSeverityLevel`), `query/index.ts` (`leveledBySeverity`), `view.ts` (`formatAlertSeverity`), and `audit-helpers.ts` (`asSeverityLevel`).

5. **CISA KEV exemption from `--audit-level`** — Actively exploited findings survive severity filtering. Tested in `audit-behaviour.ts`. Good security design.

6. **`action: 'ignore'` filtering** — Filtered at the single entry point (`getPackageAlerts`) so counts and rendering can't disagree. Tested with fixtures that demonstrate a package dropping out entirely when all alerts are ignored.

7. **`PackageAlert` type enrichment in `security-archive`** — `AlertType`, `AlertCategory`, `AlertSeverity`, `AlertAction`, `AlertFix`, `AlertPatch` — comprehensive typing of the feed's data model, with `(string & {})` fallback for unknown values. Clean.

#### Observations (non-blocking)

1. **`isSecuritySelector` vs `isSecurityAuditSelector`** — Two regex-based selector detectors. The distinction is documented (`:scripts` is security-relevant for `query` but not for `audit`), but it's regex-matching on DSS query strings rather than parsed AST. For now this works because the selector names are unique enough, but if a selector named `:malware-info` were added, the regex would false-positive.

2. **`needsRegistry = true` on `audit.ts`** — Correct for audit (it always hits the security archive), but worth noting in a comment that this is intentional, since the previous review flagged `needsRegistry` on query.

3. **Synthetic alert round-trip** — The AUDIT.md documents this as load-bearing for `vlt query`'s footer. It's a reasonable trade-off for now, but it means `vlt query :malware` (without an archive) and `vlt audit` can disagree on alert details since one synthesizes from insights and the other reads the archive.

4. **`scanCoverage` walks `graph.nodes` independently** — It counts nodes the archive has vs doesn't have data for. This is correct and matches the design doc's intent ("coverage is measured over the whole graph, not the results").

---

### Summary

The PR has matured significantly. The previous critical issue (hyperlink escapes) is fixed. The code is well-documented via AUDIT.md, has proper terminal safety via `safe-text.ts`, clean command/view separation, and thorough test coverage including integration tests with the real DSS engine. The main remaining consolidation opportunity is the alert-type set duplication across packages.

**Verdict: Approve-quality** with the set duplication noted as a follow-up.

src/query/src/index.ts
```diff
@@ -40,6 +44,115 @@ export type SearchOptions = {
 
 const noopFn = async (state: ParserState) => state
 
+const malwareAlertTypes = new Set(['malware', 'gptMalware'])
+
+/**
+ * The level to assume for an alert type when the alert's own severity
+ * is missing or unrecognized.
+ *
+ * Several of these type names encode a grade -- `mildCVE`, `mediumCVE`,
```
vltbaudbot, line 53: **Consolidation opportunity:** `cveAlertTypes` is defined identically here and in `src/query/src/pseudo/severity.ts:57`. `malwareAlertTypes` is defined identically here and in `src/cli-sdk/src/audit-helpers.ts:200`. `typeImpliedLevel` is the same data as the `Map` in `severity.ts:72` but as a `Record`.

These could live in `@vltpkg/security-archive` (which already exports `isVulnerabilityAlert` and `VulnerabilityAlertType`) and be imported by both `@vltpkg/query` and `@vltpkg/cli-sdk`. Not blocking, but the 3–4× duplication will drift as new alert types are added to the feed.

Files:
- `src/query/src/index.ts:47` — `malwareAlertTypes`
- `src/query/src/index.ts:59` — `typeImpliedLevel` (Record)
- `src/query/src/index.ts:127` — `cveAlertTypes`
- `src/query/src/index.ts:140` — `vulnAlertTypes`
- `src/query/src/pseudo/severity.ts:57` — `cveAlertTypes` (duplicate)
- `src/query/src/pseudo/severity.ts:72` — `typeImpliedLevel` (Map, duplicate)
- `src/security-archive/src/types.ts:243` — `vulnerabilityAlertTypes` (superset)
- `src/cli-sdk/src/audit-helpers.ts:200` — `malwareAlertTypes` (duplicate)

vltbaudbot (commented):
Inline note on the main consolidation opportunity.

src/cli-sdk/AUDIT.md
usrrname, line 1: This is more like a requirements/design doc. usually people keep it in a `/prd` or `docs/` folder in agentic codebases. It could be a `README`

src/cli-sdk/src/commands/audit.ts
```diff
@@ -0,0 +1,154 @@
...
+  // Coverage is measured over the whole graph, not the query results --
+  // every result matched a security selector and so had feed data, which
+  // would report near-total coverage regardless of the truth.
+  const coverage = scanCoverage(graph.nodes.values())
+  result.scannedCount = coverage.scanned
+  result.unscannedCount = coverage.unscanned
+
+  const filtered = filterAuditResult(result, auditLevel)
+
+  // Exit with error code when findings are present, matching npm/pnpm behavior
```
ruyadorno:
```suggestion
  // Exit with error code when findings are present
```

src/cli-sdk/AUDIT.md
ruyadorno: I think we should move it to `.cursor/rules/` which is the place we have these in this repo

ruyadorno:
It turned out that the `security-archive` is broken when using the default setup provided by `vlt setup` (so that was the reason why I was missing results in my local machine) - I'm adressing that problem in https://github.com/vltpkg/vltpkg/pull/1773 - but that little setup problem aside, the audit command looks great now 👍 

I think we just need to clear these last findings from the bot and it should be good to go.

src/query/src/index.ts
usrrname: This was completed in https://github.com/vltpkg/vltpkg/pull/1769

usrrname:
Re: 3. sqlite cache reads aren't registry-filtered (masks issue 1 nondeterministically)
Defer to after #1773 merged. 
 
#1773 already changes `#queueUpRequiredPackages` to use `usesNpmRegistry`. If #1773 doesn't also filter `#loadItemsFromDatabase` reads, it should be addressed there or in a follow-up after merge, so the fix uses the correct predicate. Applying it here would create a merge conflict. cc @ruyadorno

ruyadorno:
> Re: 3. sqlite cache reads aren't registry-filtered (masks issue 1 nondeterministically) Defer to after #1773 merged.
> 
> #1773 already changes `#queueUpRequiredPackages` to use `usesNpmRegistry`. If #1773 doesn't also filter `#loadItemsFromDatabase` reads, it should be addressed there or in a follow-up after merge, so the fix uses the correct predicate. Applying it here would create a merge conflict. cc @ruyadorno

Sounds good 👍 I don't think we necessarily need to change anything but no need to do anything here, I can follow up in the future if we observe actual problems

usrrname:
@ruyadorno First 2 points from RuyBot were updated with TDD and I rebased as my opencode GLM 5.2 allowed me to, put it to Draft for now

PR #1777, opened 2026-08-19
fix(rollback-remove): don't register a path as backed up until rename succeeds

### LINKED ISSUE #1776: [BUG] RollbackRemove.rm() leaves a stale entry on non-ENOENT/EPERM rename failure — rollback() can delete un-backed-up data
### ✅ Expected Behavior

A failed `rm()` should never leave `RollbackRemove` believing a backup exists when it doesn't. `rollback()` should be a safe no-op for any path whose backup rename never actually happened — it should never delete the original file when no backup of it exists.

### ❌ Actual Behavior

`RollbackRemove.rm()` (`src/rollback-remove/src/index.ts`) registers `path → target` in its internal map *before* the `rename()` that's supposed to back the file up has succeeded. Only `ENOENT`/`EPERM` failures un-register that entry — any other rename failure (realistically `EBUSY`, common on Windows when antivirus or another process holds a handle on something under `node_modules`) leaves the map claiming a backup exists at `target`, when `path` was never actually moved.

That map is exactly what `rollback()` trusts during a failed install (`src/graph/src/reify/index.ts` → `rollback()` → `remover.rollback()`):

```ts
rimraf(original)                        // deletes the ORIGINAL, never-moved file
  .catch(() => {})
  .then(() => rename(moved, original))  // fails — `moved` never existed
```

So on a real install failure: a package's removal-rename hits `EBUSY` → the map still thinks it's backed up → install fails later for an unrelated reason → the `finally` block's rollback runs → it deletes the real package directory that was never moved, then tries to restore from a backup that doesn't exist. That second failure is silently swallowed by `.catch(() => {})` at two call sites, so the user just sees "install failed" with no indication a package was permanently deleted.

Confirmed this is distinct from #661 (closed) — that was a leftover-marker cleanup bug from EPERM handling, unrelated to this data-loss path.

### 🔄 Steps to Reproduce

1. Trigger a `reify()`/install operation that removes or replaces an existing package under `node_modules/.vlt`
2. Have the `rename()` call inside `RollbackRemove.rm()` for that package fail with an error other than `ENOENT`/`EPERM` — in practice this happens on Windows when another process (antivirus, an editor, a running script) holds an open file handle inside that package's directory, producing `EBUSY`
3. Have any later step in the same `reify()` call fail for an unrelated reason, triggering the `finally` block's rollback path
4. Observe: the package directory from step 2 is deleted by `rollback()`'s `rimraf(original)` call, with no backup to restore from, and no error surfaced to the user — the install just reports failure

A regression test reproducing this deterministically (mocking `rename` to throw `EBUSY`) is attached in the linked PR.

### 🌍 Environment

- OS: win32 v24.15.0
- Node: v24.15.0 (i also have v22.23.2 i checkout in between)
- vlt: 1.0.3

### 🔄 Frequency

Occasionally - Happens sometimes

drawliin:
Fixes #1776 

Closes the data-loss bug in #1776 : `RollbackRemove.rm()` registered a `path → target` mapping *before* confirming the `rename()` that's supposed to back the file up actually succeeded. Only `ENOENT`/`EPERM` failures un-registered that entry.. any other rename failure (realistically `EBUSY`, common on Windows when antivirus or another process holds a handle on something under `node_modules`) left a stale entry claiming a backup existed when the original file was never moved.

That map is what `rollback()` trusts during a failed install. With a stale entry, `rollback()` would `rimraf(original)` believing it had a backup to restore from, permanently destroying data that was never actually backed up.. then fail to restore it (since the "backup" never existed), with that failure silently swallowed by `.catch(() => {})` at two call sites.

### The fix

Only register the mapping after `rename()` succeeds, instead of before:

```ts
try {
  await rename(path, target)
} catch (e: unknown) {
  if (e instanceof Error && 'code' in e && (e.code === 'ENOENT' || e.code === 'EPERM')) {
    return
  }
  throw e
}
this.#paths.set(path, target)
```

This closes the whole bug class in one change rather than trying to enumerate every OS error code that can legitimately fail a same-directory rename.

### Test plan

- [x] New regression test in `src/rollback-remove/test/index.ts`.. mocks `node:fs/promises` (spreading the real module and overriding only `rename`, since replacing the whole namespace broke other consumers needing e.g. `readlink`) to fail with `EBUSY`, then asserts the original file survives `rollback()` untouched
- [x] `npx tap test/index.ts` in `src/rollback-remove`.. 16/16 pass
- [x] Full package suite (`npx tap`).. 19/19 pass, no regressions
- [x] Verified on Windows (Node v22.23.2)

### DIFF
--- a/src/rollback-remove/src/index.ts
+++ b/src/rollback-remove/src/index.ts
@@ -14,8 +14,10 @@ export class RollbackRemove {
   async rm(path: string) {
     if (this.#paths.has(path)) return
     const target = `${dirname(path)}/.VLT.DELETE.${this.#key}.${basename(path)}`
-    this.#paths.set(path, target)
-    await rename(path, target).catch((e: unknown) => {
+    // rollback() trusts this map, so only record entries after rename succeeds.
+    try {
+      await rename(path, target)
+    } catch (e: unknown) {
       if (
         e instanceof Error &&
         'code' in e &&
@@ -23,12 +25,11 @@ export class RollbackRemove {
         (e.code === 'ENOENT' || e.code === 'EPERM')
         /* c8 ignore stop */
       ) {
-        this.#paths.delete(path)
         return
       }
-      /* c8 ignore next */
       throw e
-    })
+    }
+    this.#paths.set(path, target)
   }
 
   confirm() {
--- a/src/rollback-remove/test/index.ts
+++ b/src/rollback-remove/test/index.ts
@@ -1,6 +1,7 @@
 import assert from 'node:assert'
 import type { SpawnOptions } from 'node:child_process'
-import { readdirSync } from 'node:fs'
+import { readdirSync, readFileSync } from 'node:fs'
+import * as fsPromisesOrig from 'node:fs/promises'
 import t from 'tap'
 import type { Test } from 'tap'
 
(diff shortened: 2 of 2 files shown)


ruyadorno (approved):
Thank you for the contribution! 👍 LGTM

PR #1786, opened 2026-08-21
fix(graph): dedup deps listed in both dependencies and optionalDependencies

vltbaudbot:
## Summary

When a package lists the same dependency in both `dependencies` and `optionalDependencies` (e.g. `esbuild@0.18.20`, `typescript@7.0.2`), vlt creates the node as non-optional from the `dependencies` entry, then the `optionalDependencies` entry cannot recover the `optional` flag. This causes platform-specific binaries for all platforms to be materialized, wasting ~728 MB (40% of total install size in vlt.io).

## Root Cause

In `src/graph/src/ideal/append-nodes.ts`, the dep collection in `processPlacementTasks` iterates `longDependencyTypes` (which includes `dependencies` before `optionalDependencies`) and pushes entries into a plain `Dependency[]` array without deduplicating by name. When both fields contain the same package name:

1. The `dependencies` entry is pushed first with `type='prod'`
2. The `optionalDependencies` entry is pushed second with `type='optional'`
3. In `placePackage`, the first entry creates the node with `optional=false`
4. The second entry finds the existing node, but `toFoundNode.optional &&= flags.optional` evaluates to `false && true = false` — the flag can never be set back to `true`

Result: `node.isOptional() === false` → `extractNode` skips the `platformCheck` → all platform variants are materialized.

Per npm semantics, entries in `optionalDependencies` should override entries of the same name in `dependencies`.

## Fix

Changed the dep collection from a plain array to a `Map<string, Dependency>` keyed by name. Since `longDependencyTypes` iterates `optionalDependencies` after `dependencies`, `Map.set` naturally gives `optionalDependencies` precedence. This is the same dedup pattern already used by `getRawDependencies()` in `dependencies.ts` for importers.

## Affected Packages

Packages using the legacy npm pattern of listing platform deps in both fields:
- `esbuild@0.18.20` — 22 platform packages × ~10 MB each (newer esbuild versions fixed this on the publisher side)
- `typescript@7.0.2` — 20 platform packages × ~30 MB each
- Any future package using the same pattern

## Tests

Added two regression tests:
1. **Direct case**: parent with overlapping `dependencies`/`optionalDependencies` — verifies the child node gets `optional=true` and only one edge exists
2. **Deep transitive chain**: same scenario but through a 3-level dep chain (simulating the `drizzle-kit → @esbuild-kit → esbuild` path) — verifies the optional flag propagates correctly through deep chains

### DIFF
--- a/src/graph/src/ideal/append-nodes.ts
+++ b/src/graph/src/ideal/append-nodes.ts
@@ -590,7 +590,15 @@ const processPlacementTasks = async (
     )
 
     // setup next level to process all child dependencies in the manifest
-    const nextDeps: Dependency[] = []
+    // Use a Map keyed by name so that later dependency types override
+    // earlier ones. This matches npm semantics where entries in
+    // `optionalDependencies` override entries of the same name in
+    // `dependencies`. Without this, a package listed in both
+    // (e.g. esbuild@0.18.20, typescript@7.0.2) would first be placed
+    // as a regular (non-optional) dep, and the later optional entry
+    // could not recover the `optional` flag due to the `&&=` operator
+    // in `placePackage`.
+    const nextDepsMap = new Map<string, Dependency>()
 
     // traverse actual dependency declarations in the manifest
     // creating dependency entries for them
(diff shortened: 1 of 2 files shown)


ruyadorno (approved):
LGTM 👍

PR #1734, opened 2026-08-10
feat: markdown content negotiation, adds "Copy as markdown" button on each page; 

usrrname:
## Summary

The purpose of this PR is to:

1. Enable markdown content negotiation for AI agents 
2. provide a copy-markdown button for developers browsing the documentation.
3. advertise agent-skill via Link response header on /
Adds a `Link: rel="agent-skill"` header on the homepage pointing to the
markdown-negotiation `SKILL.md`, per RFC 8288, so agent-discovery
scanners find it without crawling `.well-known` directly. Applied at
Vercel's edge/headers layer since the homepage is static.

## Note

Output Directory is misconfigured, needs a fix that touches prod

  Found while debugging 404s on the new `/api/markdown/:slug` route:
  the project's Output Directory is pinned to
  `www/docs/.vercel/output/static`. Since `www/docs` uses SSR
  (`@astrojs/vercel adapter`), that setting makes Vercel deploy only
  static assets and silently drop the serverless function + routing
  config — so any dynamic/SSR route 404s. Static pages are
  unaffected, which is why it's gone unnoticed.

  This is a project-level setting, so clearing it affects prod too
  (not just previews). Should be a no-op for existing static pages,
  but flagging before I flip it — shout if you know why it was set
  this way. Plan: clear it, verify prod still loads, ship.

## Changes

- Custom API route at `src/pages/api/markdown/[...slug].ts` that converts
  rendered docs to markdown using:
    - Astro's experimental `AstroContainer` to render pages server-side in `lib/markdown.ts`
    - Turndown to convert HTML to Markdown
    - GitHub-flavored markdown support via `@joplin/turndown-plugin-gfm`
  - Route works perfectly in local development
  - Added `export const prerender = false` at `src/pages/api/markdown/[...slug].ts` to  to keep it dynamic

### 1. Copy-to-Markdown Button
- One-click button to copy documentation as markdown to clipboard
- Available on every doc page
- Shows loading/success/error states with visual feedback
- Comprehensive ARIA accessibility features

### 2. Markdown API Endpoint
- `GET /api/markdown/<slug>` returns any doc page as GitHub-flavored markdown
- Converts rendered HTML to markdown with proper formatting
- Handles code blocks, tables, links, and images
- Graceful error handling with markdown-formatted error messages

### How It Works

1. **User clicks button** → Fetches markdown via `/api/markdown/<slug>` → Copies to clipboard
[...]

PR #1789, opened 2026-08-21
fix: guard ignoreCommand against orphaned VERCEL_GIT_PREVIOUS_SHA

usrrname:
## Summary

- The `ignoreCommand` in `vercel.json` was originally added to skip Vercel builds when a commit doesn't touch `www/docs/`, preventing excessive builds on every merge to `main`.
- On feature branches, interactive rebases (or force-pushes) can leave `$VERCEL_GIT_PREVIOUS_SHA` pointing at a commit that no longer exists in history. `git diff` against a missing SHA fails fatally (`fatal: bad object`), which broke the build instead of just skipping it.

Screenshot 2026-08-21 at 5 34 10 PM

As a workaround that _SOMETIMES_ worked everytime I updated a branch, I kept having to uncheck "ignore build" checkbox

- The command now checks that `$VERCEL_GIT_PREVIOUS_SHA` exists with `git cat-file -e` first, and falls back to always building when it doesn't — preserving the original intent (skip unnecessary docs builds) without crashing on rebased/force-pushed branches.

## Test plan

- [x] Confirm `vercel.json` is valid JSON
- [ ] Verify a normal merge to `main` still skips the build when `www/docs/` is untouched
- [x] Verify a feature branch with a rebased/orphaned previous SHA no longer fails the build (was tested on #1734's branch)

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://openapi.vercel.sh/vercel.json",
-  "ignoreCommand": "git diff --quiet $VERCEL_GIT_PREVIOUS_SHA $VERCEL_GIT_COMMIT_SHA -- www/docs/",
+  "ignoreCommand": "git cat-file -e \"$VERCEL_GIT_PREVIOUS_SHA\" 2>/dev/null && git diff --quiet \"$VERCEL_GIT_PREVIOUS_SHA\" \"$VERCEL_GIT_COMMIT_SHA\" -- www/docs/ || exit 1",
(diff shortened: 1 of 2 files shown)


usrrname:
Also fixed a broken import in \`dashboard.mdx\` (\`../../components/image/image.astro\` → \`../../../components/image/image.astro\`) that was failing the Vercel build independently of the \`ignoreCommand\` fix. Verified locally with \`vlr --workspace ./www/docs build\` — 188 pages built, all internal links valid.

ruyadorno (approved):

PR #1793, opened 2026-08-22
fix(tar): delete completed requests from Pool.pending

### LINKED ISSUE #1792: [BUG] Pool.pending never removes completed requests — memory leak holding every unpacked tarball buffer
### ✅ Expected Behavior

Once an unpack request resolves or rejects, `Pool` should stop holding a reference to it. Memory usage from the unpack pool shouldn't grow proportionally with total install size and never come back down.

### ❌ Actual Behavior

`Pool#onMessage()` (`src/tar/src/pool.ts`) retrieves a completed request from `this.pending` to resolve/reject it, but never calls `this.pending.delete(id)`. Entries are only ever added (`unpack()`), never removed.

`UnpackRequest` (`src/tar/src/unpack-request.ts`) holds `tarData: Buffer`.. the full raw tarball for that package. So every package unpacked during the life of the process stays referenced in `pending` forever, even after the install finishes successfully. For a large install, that's the sum of every package's tarball size pinned in the heap permanently.

### 🔄 Steps to Reproduce

1. Run an install with many dependencies (or call `Pool.unpack()` directly in a loop)
2. After all unpacks resolve, inspect `pool.pending.size`. it will equal the total number of unpacks performed, not 0
3. Each retained entry holds a full tarball `Buffer`, not just a small handle

### 🌍 Environment

- OS: linux v22.23.2
- Node: v22.23.2
- vlt: 1.0.4

### 🔄 Frequency

Always - Happens every time

drawliin:
Closes #1792

`Pool#onMessage()` resolved/rejected a completed `UnpackRequest` but never removed it from `this.pending`, so every unpacked tarball's buffer stayed referenced in memory for the life of the process.

### Fix: 
delete the request from pending immediately when the worker reports completion, before resolving or rejecting the promise.

### Test plan

- [x] New assertions in `src/tar/test/pool.ts` — `pending.size === 0` after both existing "unpack all the things" tests complete

### DIFF
--- a/src/tar/src/pool.ts
+++ b/src/tar/src/pool.ts
@@ -44,6 +44,8 @@ export class Pool {
     const ur = this.pending.get(id)
     /* c8 ignore next */
     if (!ur) return
+    // request is done, stop holding its tarball buffer in memory
+    this.pending.delete(id)
     if (isResponseOK(m)) {
       ur.resolve()
       /* c8 ignore start - nearly impossible in normal circumstances */
--- a/src/tar/test/pool.ts
+++ b/src/tar/test/pool.ts
@@ -88,6 +88,7 @@ t.test('unpack all the things!', async t => {
       true,
     )
   }
+  t.equal(p.pending.size, 0, 'pending must be empty after unpack')
 })
 
 t.test('unpack all the things, but flattened', async t => {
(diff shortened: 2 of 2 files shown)


ruyadorno (approved):
👍 LGTM

PR #1803, opened 2026-08-23
fix: preserve swr factor when decoding cache entries

### LINKED ISSUE #1802: [BUG] CacheEntry.decode() drops RegistryClient staleWhileRevalidateFactor on cached reads
### ✅ Expected Behavior

when `RegistryClient` is configured with a custom `staleWhileRevalidateFactor`, cached reads should preserve that same factor after decoding a stored cache entry..

### ❌ Actual Behavior

fresh writes pass the configured factor into `CacheEntry`, but cached reads go through `CacheEntry.decode()` without passing it back in.. decoded entries then fall back to `CacheEntry`'s class default of `60` instead of the `RegistryClient` value..

in practice, that means the configured stale while revalidate window is not what actually gets enforced on reads from disk..

### 🔄 Steps to Reproduce

1. construct a `RegistryClient` with a non-default `staleWhileRevalidateFactor`..
2. fetch and cache a response that relies on the fallback swr behavior..
3. read the same cached entry back through the normal cached-read path..
4. observe that the decoded `CacheEntry` behaves as if its factor were `60` rather than the configured client value..

### 🌍 Environment

- OS: linux 22.22.3
- Node: 22.22.3
- vlt: 1.0.5

### 🔄 Frequency

Always - Happens every time

drawliin:
closes #1802

## summary
fixes the cached-read path so `RegistryClient` preserves its configured `staleWhileRevalidateFactor` when decoding cache entries..

## testing
- npm test -- test/cache-entry.ts
- npm run typecheck

### DIFF
--- a/src/registry-client/src/cache-entry.ts
+++ b/src/registry-client/src/cache-entry.ts
@@ -623,7 +623,10 @@ export class CacheEntry {
    * and this static method will decode it into a CacheEntry representing
    * the cached response.
    */
-  static decode(buffer: Uint8Array): CacheEntry {
+  static decode(
+    buffer: Uint8Array,
+    options: CacheEntryOptions = {},
+  ): CacheEntry {
     const parsed = CacheEntry.#parseHead(buffer)
     if (!parsed) return emptyCacheEntry
     const body = buffer.subarray(parsed.headSize)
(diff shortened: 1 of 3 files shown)


ruyadorno (approved):
LGTM! 👍

PR #1804, opened 2026-08-23
fix: continue cache revalidation after empty entries

### LINKED ISSUE #1801: [BUG] cache-revalidate beforeExit exits early when a registered path has an empty set
### ✅ Expected Behavior

`handleBeforeExit()` in `cache-revalidate.ts` should keep iterating registered cache paths even if one entry has no pending urls to revalidate..

### ❌ Actual Behavior

the loop currently does `if (!r.size) return`..

if an empty set is encountered, the handler returns immediately and skips revalidation for every later registered path..

### 🔄 Steps to Reproduce

1. register more than one cache path for background revalidation..
2. ensure one of the stored path entries has an empty url set by the time `beforeExit` runs..
3. trigger the `beforeExit` handler..
4. observe that later pending paths are never processed..

### 🌍 Environment

- OS: linux 22.22.3
- Node: 22.22.3
- vlt: 1.0.5

### 🔄 Frequency

Always - Happens every time

drawliin:
closes #1801

## summary
changes `handleBeforeExit()` to skip empty registered sets instead of returning early.. adds a regression test proving later pending paths still get processed..

## testing
- npm test -- test/cache-revalidate.ts
- npm run typecheck

### DIFF
--- a/src/registry-client/src/cache-revalidate.ts
+++ b/src/registry-client/src/cache-revalidate.ts
@@ -30,7 +30,7 @@ const handleBeforeExit = () => {
   const compileCacheDir = module.getCompileCacheDir()
   for (const [path, r] of registered) {
     /* c8 ignore next */
-    if (!r.size) return
+    if (!r.size) continue
     const env = { ...process.env }
     if (compileCacheDir) env.NODE_COMPILE_CACHE ??= compileCacheDir
     const args = []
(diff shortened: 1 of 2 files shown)


ruyadorno (approved):
LGTM 👍 thanks for the contribution!

PR #1805, opened 2026-08-23
feat(cli-sdk): add `vlt create pkg-example` command to clone "Publish your first package" demo package for vlt users

usrrname:
## Summary

Implements feedback from demo https://github.com/vltpkg/vlt.io/issues/1889  to use `vlt create` as a CLI-based way for beginners to retrieve a publishable demo package for the ["publish your first package"]() tutorial.

- Adds `vlt create pkg-example [target-dir]`, a special-cased path in `vlt create` that scaffolds the `packages/vlt` example from [`vltpkg/package-examples/packages/vlt`](https://github.com/vltpkg/package-examples/tree/main/packages/vlt) into a target directory.

- Fixed a registry-gating bug found during verification: `create.ts`'s static `needsRegistry = true` would have blocked `pkg-example` for any user without a configured registry, before `command()` ever ran. Moved the check inline (after the `pkg-example` early return, following the `login.ts` precedent) so the generic `create-*` flow still requires a registry but `pkg-example` doesn't.
- Cleaned up `createPkgExample`'s scratch-dir handling (removed a redundant nested subdirectory — `unpack()` already tolerates its target existing) and deduped a repeated test fixture across `create.ts`'s test file.

## Test plan
- [x] `vlr test` — full `src/cli-sdk` suite green (2247/2248, 1 pre-existing unrelated todo), including `posttest` typecheck
- [x] `vlr format` / `vlr lint` — clean
- [x] Manual smoke test against the real `vltpkg/package-examples` repo: correct scaffold, `@YOUR-ACCOUNT` placeholder intact, conflict-check rejects non-empty target dirs, both `json` and `human` view output verified

### DIFF
--- a/src/cli-sdk/package.json
+++ b/src/cli-sdk/package.json
@@ -28,7 +28,6 @@
     "@vltpkg/run": "workspace:*",
     "@vltpkg/security-archive": "workspace:*",
     "@vltpkg/spec": "workspace:*",
-    "@vltpkg/tar": "workspace:*",
     "@vltpkg/types": "workspace:*",
     "@vltpkg/url-open": "workspace:*",
     "@vltpkg/vlt-json": "workspace:*",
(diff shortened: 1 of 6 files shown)


ruyadorno:
Let's add support to the new `--save-config` option here so that the "create a new project" story is solid when using the vlt registry ⚡

PR #1814, opened 2026-08-27
fix(cli) project vlt.json settings override user-level vlt.json settings

### LINKED ISSUE #1736: [BUG] local registry project config doesn't overidde global config
### ✅ Expected Behavior

I'd expect a local `vlt.json` to override any registry configs set at the user/system level so that commands like `vlt setup` / `vlt login` / `vlt registry ...` would resolve to the relevant, local project-specific config.

### ❌ Actual Behavior

`vlt` pulls from the user/globally defined registry config forcing you to change/manage only a single set of registry configs on the system

### 🔄 Steps to Reproduce

_No response_

### 🌍 Environment

Run the following in your terminal then copy+paste the output: 
  node -p "[
    '- OS: ' + process.platform + ' ' + process.version,
    '- Node: ' + process.version,
    '- vlt: ' + require('child_process').execSync('vlt -v').toString().trim()
  ].join('\n')"

### 🔄 Frequency

Always - Happens every time

evert:
This PR ensures that project-level settings take precedence over
settings in the XDG vlt.json.

The settings objects are deeply merged, and `null` can be used on the
project level to explicitly clear a user-level one.

Lastly this adds a `--config` toggle on the setup command.

Fixes #1736

### DIFF
--- a/src/cli-sdk/src/commands/login.ts
+++ b/src/cli-sdk/src/commands/login.ts
@@ -1,4 +1,5 @@
 import { RegistryClient } from '@vltpkg/registry-client'
+import { configWriteTarget } from '../config/index.ts'
 import { commandUsage } from '../config/usage.ts'
 import { missingRegistryError } from '../require-registry.ts'
 import type { CommandFn, CommandUsage } from '../index.ts'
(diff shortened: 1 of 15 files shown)


ruyadorno (approved):
LGTM 👍

PR #1379, opened 2025-12-19
fix(graph): preserve peer dep target when forking context

lukekarrys:
## Summary

- Fixes peer dependency resolution to preserve pinned versions consistently across all dependent packages
- When a project specifies a pinned peer dep (e.g., `typescript@5.7.3`), all packages with compatible peer dep ranges now resolve to that same version
- Prevents later-resolved packages from overwriting earlier targets when the existing target still satisfies the spec

## Problem

When a project has:
- `typescript@5.7.3` (pinned) as a direct dependency  
- `typescript-eslint@^8.49.0` which has peer deps on TypeScript

vlt was incorrectly resolving some `@typescript-eslint/*` packages to `typescript@5.9.3` while others got `5.7.3`, causing ESLint parsing errors due to AST incompatibilities between TypeScript versions.

## Solution

Three changes to peer dependency resolution in `src/graph/src/ideal/peers.ts`:

1. **`resolvePeerDeps` now accepts the current context**: Previously captured context at closure creation, but contexts may be forked before resolution runs

2. **`addEntriesToPeerContext` preserves existing satisfying targets**: Before updating edges to a new target, check if the existing target satisfies the new spec - if so, keep it

3. **`forkPeerContext` preserves parent targets**: When new entries have no target but parent context does, preserve the parent's resolved target

Fixes #1378

PR #1815, opened 2026-08-28
graph: improve lockfile churn

### LINKED ISSUE #1465: [BUG] improve peer context refs in lockfile
Right now updating an install graph in a large project causes a large amount of churn in the lockfile due to varying indexes.

See https://github.com/vltpkg/vltpkg/pull/1450 for an example of graph update that causes thousands of lines changes in the lockfile.

ruyadorno:
Peer nodes used a non-deterministic `peer.N` suffix in their DepIDs, causing needless lockfile rewrites while also dropping integrity/resolved data.

This changeset canonicalize peer IDs via content hashing, preserving prior edge resolutions across rebuilds, and fixing security-archive lookups to use base DepIDs.

Fixes: https://github.com/vltpkg/vltpkg/issues/1465
Refs: https://github.com/vltpkg/vltpkg/issues/1404

## Locked version hydration (no `node_modules`)

Follow-up commits `1491c60b` + `83f816b7`. Rebuilding with a lockfile but no `node_modules` (fresh clone + `vlt i newpkg`, or `vlr wipe && vlt i x`) re-resolved every dependency to the newest satisfying version: lockfile nodes only get a manifest from the hidden lockfile or the actual graph, so the locked target stayed detached, `checkPeerEdgesCompatible` saw no peer deps and the range fetch drifted.

Now `fetchManifestsForDeps` first hydrates every fitting locked node: rebuild the exact spec from the DepID (`hydrate()` + re-parse with current options, registry must match the edge's), fetch it once per node, set `node.manifest`. The rest of the pipeline then behaves exactly like the hidden-lockfile path. Any miss (unpublished version, registry moved, fetch error) leaves the node bare and the range fetch runs as before. No extra network: same packument, exact-spec manifest cache entry.

Also fixes `loadNodes` never parsing a version for scoped names (`indexOf('@') > 0` is 0 for `@scope/x@1.2.3`), which hid this path for every scoped package.

Validation on vlt.io (`stress/s1c`, `lockdiff.mjs` from the churn report):

| check | before | after |
| --- | --- | --- |
| wipe + `vlt i abbrev` vs same add with `node_modules` | +56 / −48 nodes, 23 renames | 0 diff lines |
| fresh-clone `vlt i abbrev --lockfile-only` vs same | same drift | 0 diff lines |
| noop `vlt i` after wipe + add | 0 | 0 |
| wipe + add wall time (3 runs, warm cache) | 40.8 / 40.7 / 37.7 s | 35.1 / 33.1 / 35.9 s |

## Peer-context rewiring (`c302c777`, `305adafe`, `a0415881`, `daef9514`)

Every zero-dep `vlt i <pkg>` on vlt.io flipped recharts' peer edge `react-is` between 16.13.1 and 17.0.2 (17 lockfile lines per add), and recharts' prod `use-sync-external-store` landed on the copy whose peer react (19.2.8) differs from recharts' own (19.2.5). Three causes, three commits:

1. **`forkPeerContext` copied `contextDependents`.** Inherited entries start with `target: undefined` and the *parent's* dependents, so the first fork-local placement of that name ran the swap loop over nodes living in the parent context. Whether it fired depended on what the previous lockfile locked, so the graph alternated between two fixpoints. Inherited entries now start with no dependents; specs are still copied (they gate `incompatibleSpecs` and the fork key).

2. **`checkPeerEdgesCompatible` counted devDependencies of non-importers.** `shouldIgnoreContextMismatch` and CHECK 3 scanned every dependency type for a declaration of the peer and returned on the first hit. Registry manifests carry devDependencies, so any package pinning a peer there (recharts, most React libs) got the "parent resolves its own copy" escape hatch for a dep that is never installed: CHECK 1 was skipped and a copy with a different peer was reused. Both loops now mirror placement via `shouldInstallDepType` (moved to `dependencies.ts`, shared with `append-nodes.ts`): devDependencies count for importers and `git:` deps only. `file:` follows placement, not `getRawDependencies`, since a `file:` dep's devDependencies are never placed in the ideal graph.

3. **The swap re-pointed dependents that were already satisfied.** A dependent is now re-pointed only when its current target fails a spec the entry holds, i.e. when the context genuinely outgrew it. The `id !== && version !==` gate stays: two peer copies of one version are the same resolution for the swap, and moving a dependent between them would only change its peer set. The incoming spec is collected before the swap runs (`daef9514`): intersecting ranges do not fork, so a dependent kept on the old target has to satisfy the new spec too.

CHECK 2's leniency (301–316) is deliberately kept; fix 2 closes the path that made it reachable here.

Validation on vlt.io (`lockdiff.mjs`, source-run `vlt`):

| step | before | after |
| --- | --- | --- |
| `stress/s1b` (react-is 16.13.1) + 3 zero-dep adds | flips 17/16 lines every add | 1 settle to 17.0.2 (18/16), then 1/0 lines, 0 renames |
| `stress/s1c` + 2 zero-dep adds | 17/16 lines every add | 1 settle (`use-sync-external-store` → the react-19.2.5 copy), then 1/0, 0 renames |
| `stress/base` → add `abbrev` to `packages/shutdown` | `use-sync-external-store` moves to the 19.2.8 copy | stays on `~peer.9aa5b9cd9e13b174`; 142/204 lines, 5 env renames = 4 peerOptional `MISSING`→resolved + the react-is settle |
| noop `vlt i` / `--lockfile-only` add / wipe + add | — | 0 lines / byte-identical to the `node_modules`-backed add |

Noop and `--lockfile-only` wall time is unchanged (3 interleaved rounds, medians). The five gold peer snapshots pass untouched. Two checked-in `build()` regressions cover the cycle and the divergence, both red on `83f816b7`.

Out of scope, unchanged: the remaining peerOptional `MISSING`↔resolved flips, and the `@sentry/react` / `@vercel/microfrontends` moves in the `next@latest` step, which are a cascade of a `next` downgrade rather than the swap.

## Dist-tag resolution and the saved spec (`dec39fd1`, `79708a9b`, `92f172bb`, `4d22ecb5`)

`vlt i next@latest` in vlt.io's `apps/www` **downgraded** next 16.2.6 → 16.1.7 and wrote `prod latest` next to `"next": "^16.1.7"`, so `--frozen-lockfile` failed forever after and plain/add installs alternated between two graphs (±40 nodes). Four commits:

1. **Explicit dist-tag adds resolve the tag.** A dist-tag satisfies *any* version of that name, so the locked node "fits", `checkPeerEdgesCompatible` CHECK 0 rejects it (detached after `resetEdges`), and the candidate fallback then takes whichever same-name copy happens to be attached. A name that came from the command line now bypasses reuse and fetches the tag. The signal is a snapshot of `options.add` taken *before* `buildIdealFromStartingGraph` merges manifest-derived deltas into it — `add`, `importerSpecs.add` and `transientAdd` all carry entries the user never typed. Transitive `"foo": "latest"` and lockfile edges keep today's reuse (npm keeps the lenient rule there too); a registry spec naming a workspace still links the workspace. Explicit adds also stop swallowing resolution failures: `vlt i --save-optional foo@nosuchtag` used to exit 0 having written nothing, or blow up later with `Dependency node could not be found`.

2. **The importer edge stores the saved value.** The lockfile wrote `edge.spec.bareSpec` while package.json got `calculateSaveValue(...)`. `addOrRemoveDeps` now writes that value back onto the edge. That only helps if it runs: `Diff` compares edge *targets* only, so an add landing on the node already in place has no diff and `reify()` returned before `updatePackageJson` — `vlt i foo@1.2.3` and `--save-exact` wrote nothing at all, while `--lockfile-only` and any run sharing a real diff persisted the un-normalised edge. Normalisation now happens before that early return, which saves the lockfiles when the graph is flagged stale; `install.ts`'s lockfile-only path runs it before saving.

3. **Legacy lockfiles heal.** `removeSatisfiedSpecs` already finds an importer edge whose text package.json disagrees with while the locked target still satisfies it — it just dropped the text. It now reports those, `getImporterSpecs` passes them up as `staleSpecs`, and the rewrite happens before the rebuild reads `edgesOut`. Names the caller asked for are excluded (they go through 1 and 2 instead). No edge changes target. `--frozen-lockfile` / `vlt ci` stay strict: one local `vlt install` heals, commit, CI green.

4. **A rebuild-only spec change flags the lockfile.** `vlt i foo@1.2.3` over a `foo@^1.2.3` edge on the same target rewrites the edge during the ideal build, not in `updatePackageJson`, so package.json changed and the lockfile did not.

Validation on vlt.io (source-run `vlt`, `--lockfile-only`, settled `stress/s1c` + one add as the base):

| step | before | after |
| --- | --- | --- |
| `apps/www` `vlt i next@latest` | 16.1.7 (**downgrade**), edge `latest`, pj `^16.1.7`; −11/+0 nodes | `16.3.4` own copy, edge `prod ^16.3.4`, pj `^16.3.4`; +39/−10, 14 renames |
| plain `vlt i` after | +40 nodes | 0 lines |
| `--frozen-lockfile` after | `next spec changed from "next@latest" to "next@^16.1.7"` | exit 0 |
| unrelated add after | flips back, −40 | 0 real, 48 renames (one-time settle), then 0 |
| same `vlt i next@latest` again | flips | 0 lines |
| `vlt i next@canary` | 16.1.7, tag ignored | `16.4.0-canary.17`, pj `^16.4.0-canary.17` |
| `vlt i lucide-react@latest` | stays 0.562.0 | `1.41.0` |
| lockfile with `prod latest` + pj `^0.562.0`, plain `vlt i` | edge stays `latest`, `vlt ci` broken forever | exactly that one line heals, 0 nodes, frozen passes |
| `vlt i next@latest` when already at latest | nothing written; `--lockfile-only` writes `prod latest` | 0 real, edge/pj unchanged; `--save-exact` → `16.3.4` in both, 1 line |
| `vlt i --save-optional next@nosuchtag` | exit 0, `optional nosuchtag → next@16.1.7` written | exit 1 `Could not resolve`, nothing written |

Scratch project (npm registry, `abbrev` latest 5.0.0), full reify, each step followed by `--frozen-lockfile`:

| op | before | after |
| --- | --- | --- |
| `vlt i abbrev@latest` on a same target | exit 0, nothing written | pj/edge `^5.0.0`, both lockfiles re-saved |
| `vlt i abbrev@5.0.0` | nothing written | pj/edge `5.0.0`, frozen passes |
| `vlt i abbrev@latest --save-exact` | nothing written | pj/edge `5.0.0` |
| `vlt i abbrev@latest --lockfile-only` | edge `prod latest`, pj `^5.0.0`, frozen broken | edge `prod ^5.0.0` |

Rides along, all documented or npm-consistent: a same-target `vlt i foo@1.2.3` now pins package.json, `--save-exact`/`--save-prefix` apply on same-target adds, and an explicit `--save-optional` add that cannot resolve now errors.

Out of scope / known residuals:

- **Re-placing a peer-heavy package re-settles once.** `vlt i next@latest` always re-places, so running it on a graph another add had settled costs 48 suffix renames (0 real nodes, edge and package.json byte-identical); the next run is 0. Same class as the one-time settle in the section above.
- **The candidate fallback still overrides a fitting lock** (`stress/s2f`): after the heal, the shared peer-incompatible `next@16.1.7` is still re-resolved to a fresh 16.3.4 (+40) and the next add drops it again. Unchanged from before this PR; the fix is an order-independent CHECK 2b in `checkPeerEdgesCompatible`.
- `vlt i foo@^1` with a satisfying lock still reuses it (npm re-resolves any explicit request); bare `vlt i foo` keeps "install if missing"; `--frozen-lockfile` stays stricter than `npm ci`.

## Deterministic optional peers (`89da4a95`, `8385bce4`, `edc2df26`)

An optional peer resolved or dangled depending on what the previous lockfile made the rebuild reuse. Three causes:

1. **`forkPeerContext` dropped inherited targets.** A fork cleared `target` on every inherited entry, so whether a subtree's optional peers resolved depended on which fork it landed in — and that depends on reuse. `vlt i next@^16.3.4` in `apps/www` left 43 `typescript` edges under `@solana/*` MISSING which the next rebuild, and a build from scratch, resolved. The target is now inherited as a snapshot (still inactive, still dependent-less, still guarded by the satisfies check at resolution time). `#1379` did this once and was reverted by `#1391` with the `contextDependents` leak still in place; that leak is gone since `c302c777`. The fork cache key now embeds the inherited `name=target` pairs, so a base whose target moved does not hand out the fork taken before the move (red at unit level without it).

2. **`checkPeerEdgesCompatible` skipped every dangling peer edge.** A copy placed where nothing provided its optional peer was reused by parents that do provide it — three `vitest` copies collapsed into the one without `jsdom`, same for `ws`→`bufferutil`/`utf-8-validate` and `@clerk/shared`→`react`. Not churn: a build with no lockfile and no `node_modules` did it too. A dangling `peerOptional` is now incompatible when the parent provides the peer: its own edge, its **pending dependency for the level**, or a satisfying context target. The pending list, not the manifest, is the source — an explicit `vlt i --save-dev jsdom` only reaches package.json in reify, and a manifest lookup left www on the shared MISSING copy. No node is predicted for a dependency the parent has not placed yet: the fork entry is target-less and the parent's live edge resolves the peer at the end of the level, so a copy is never linked to something the parent did not install. Disjoint ranges do not fork; `keepVersion` keeps the fork on the version it forked from instead of re-resolving the range.

   The reverse (a *resolved* copy reused by a parent whose context provides nothing) is deliberately allowed: the only evidence there is negative and timing-dependent, and the shared copy carries an extra satisfying link, never a missing one. Spiked anyway — byte-identical on vlt.io.

3. **A dual-declared name was placed twice.** `@react-native/codegen` lists `@babel/core` in both `dependencies` and `peerDependencies`, `vitest` does the same for `vite`; the last writer won, so the edge type depended on when the peer resolved. The peer entry is now dropped when a regular type already claimed the name, at all three collection points (`getRawDependencies`, `getImporterSpecs`, `processPlacementTasks` after the transient merge). npm and pnpm both treat the regular declaration as authoritative; `optionalDependencies` still overrides, it is collected last.

Semantics adopted, matching pnpm exactly and npm where the provider is hoisted: **an optional peer resolves iff the placement context provides a satisfying target**. "Always MISSING" was rejected — a node sees only its own links plus the internal hoist's one-highest-copy-per-name, so every optional peer would resolve at runtime to the hoisted copy whatever its parent installs. "Resolved may never go back to MISSING" as a *lock rule* was rejected too (it keeps dead links alive after `vlt rm`); it holds as a corollary of the structural rule.

Validation on vlt.io (source-run `vlt`, `--lockfile-only`, settled `spike/rec3-base`; `peer-provisions.mjs` checks every peer edge against the rule using regular dependents only):

| step | before | after |
| --- | --- | --- |
| `apps/www` `vlt i next@^16.3.4` (first placement) | +39/−10 nodes, 43 `typescript` MISSING | +44/−10, 71 renames; 45 `typescript` + 3 `@types/node` MISSING→resolved, 4 new copies |
| unrelated add after it (the settle) | 0 real, **48 renames**, 43 env flips | 0/−1, 4 renames, 1 env (`swr`, known) |
| plain / add / rm / re-add / `--frozen-lockfile` | — | 0 / 0 / 0 / 0 / exit 0 |
| checker on the settled lockfile | `missingDirect` 5 | **0** |
| scratch (no lockfile, no `node_modules`) | `vitest` still without `jsdom` | same checker counts as the settled lockfile; diff is registry drift only |
| `vlt rm jsdom` in www, then `vlt i --save-dev jsdom@^27.4.0` | www stays on the shared MISSING copy | −36 nodes then its own copy with `jsdom` linked; the next settle equals the direct settle byte for byte |
| `stress/base` (main's lockfile) → first add | 11 renames | +4/−1, 74 renames (one-time settle) |

Full reify on a scratch monorepo (npm registry): two `eslint@10.10.0` copies, only the workspace that installs `jiti` links it — in the lockfile and in `node_modules/.vlt`; `rm jiti` returns it to the shared copy; re-adding restores it in the same run. A `file:` package dual-declaring `abbrev` gets one `prod` edge.

Noop and first-placement wall time unchanged (3 interleaved rounds each, base vs branch). The five gold peer snapshots pass untouched; graph suite 2349 asserts green at 100% coverage.

Behaviour changes to note: +4 nodes on vlt.io (`@clerk/shared`, `vite`, `vitest`, `ws` copies); `@vercel/toolbar` and `@vercel/microfrontends` no longer link `vite` under www (that link came from walking `vitest`'s peer-typed `vite` edge — pnpm gives MISSING there too); a dual-declared importer dependency changes edge type; and the first modified install after upgrading pays a one-time settle.

Out of scope / known residuals:

- **Transitive provisions** (2 edges): reuse looks one level deep, so `@clerk/shared` under a root-placed `@clerk/backend` keeps the first placement's provision when www reuses `@clerk/backend`. Deterministic across incremental and scratch builds.
- **Context population by reused subtrees**: a reused subtree never populates the peer context, so an optional peer only a sibling subtree provides moves when that sibling switches between placed and reused (4 `vite` edges when www's `vitest` copy goes away). pnpm leaves those MISSING in every case.
- **The `swr` env change on the settle**: CHECK 1 accepts an attached candidate whose react satisfies the range even though the context's react differs. Pre-existing (same `resolvedMismatch` in every lockfile before this PR); the fix is the same order-independent CHECK 2b noted above.
- **A plain install without `node_modules` cannot check peers at all**: attached lockfile nodes have no manifest, so every compatibility check returns compatible. Pre-existing; the fixes take effect on the next rebuild, which every add/remove triggers.

## Catalog and configuration changes (`d7c65826`, `e22dc2eb`, `e24e7cef`)

`main` is merged first: PR #1819's `loadObject` change (the current catalog config wins over the lockfile snapshot when parsing lockfile edges) is what already makes a catalog bump land in one install. On top of that:

1. **An options-only change is written.** A config change that moves no node -- an in-range catalog widen, a new catalog entry nothing uses yet, a registry alias nothing references -- reset edges, rebuilt to the same graph, and hit reify's no-diff early return, so the new `options` block never reached `vlt-lock.json`. Every later install reset again (3.2-3.6 s vs 2.0-2.3 s on vlt.io) and `--frozen-lockfile` failed until an unrelated change wrote the file. The reset now flags the graph stale, which is the contract reify's early return already honours.

2. **`optionsChanged` is a semantic diff.** It was a `JSON.stringify` compare, so reordering keys in `vlt.json` counted as a change. The comparison is now per entry and key-order insensitive, and the list is kept on the graph, so the frozen error names what differs instead of "Configuration options have changed (e.g. modifiers, registries, catalogs)":

   ```
   Lockfile is out of sync with package.json. Run "vlt install" to update.
     Configuration options have changed:
       catalog: oxlint "^1.59.0" -> "^1.81.0"
   ```

   URL userinfo in a printed value becomes `***`; these lines end up in CI logs.

3. **Modifier projects pass the frozen check.** `GraphModifier.maybeLoad` ran *after* the frozen block and the CLI never passes `modifiers` in options, so every project with `modifiers` in `vlt.json` failed `--frozen-lockfile` and `vlt ci` on a lockfile that was in sync, with two false detail lines: the config the check compared had no modifiers, and the edge text it compared against package.json was the override value. Modifiers are now loaded before the check and handed to it, and both the spec-text comparison and Rec #3's stale-spec heal skip dependencies a modifier governs -- a governed edge carries the modifier value by construction, and a change to that value is an options change. The frozen failure is on `main` today; the every-install rewrite of both lockfiles was a branch-only side effect of the heal.

Validation on vlt.io (source-run `vlt`, settled base, `lockdiff.mjs`; "before" = this branch with `main` merged):

| step | before | after |
| --- | --- | --- |
| `catalog.oxlint ^1.59.0 -> ^1.81.0`, `vlt i` | +20/-20 nodes, 0 renames, 8.0 s | unchanged |
| plain `vlt i` / `--frozen-lockfile` after it | 0 lines / exit 0 | unchanged |
| new catalog entry nothing uses | 0 nodes, **nothing written**; every later install resets, frozen exit 1 | +1 line; next install a 3.3 s no-op, frozen exit 0 |
| in-range widen `^1.81.0 -> ^1.80.0` | same, nothing written | options line rewritten, next install 0 lines, frozen exit 0 |
| pin `^1.80.0` | -20/+20, 0 renames | unchanged |
| `--lockfile-only` bump, then plain `vlt i` | +20/-20 then 0 lines | unchanged |
| `vlr wipe` + bump | +20/-20, 0 renames | unchanged, byte-identical to the `node_modules`-backed install |
| frozen right after the edit, before installing | exit 1, generic message | exit 1, `catalog: oxlint "^1.59.0" -> "^1.81.0"` and nothing else |
| reorder catalog keys without changing values | reset rebuild + lockfile rewritten | 2.2 s, lockfile untouched, frozen exit 0 |
| add + rm `ansi-regex` in a workspace (control) | 0 real nodes, 0 renames | unchanged |

Scratch project with one modifier (`":root > #abbrev": "2.0.0"`, package.json `^2.0.0`):

| step | before | after |
| --- | --- | --- |
| fresh `vlt i`, then `vlt i --frozen-lockfile` | exit 1: generic options line + `.: abbrev spec changed from "abbrev@2.0.0" to "abbrev@^2.0.0"`; `vlt ci` the same | exit 0, both |
| plain `vlt i` after | both lockfiles rewritten every time (content identical) | both untouched |
| modifier value `2.0.0 -> 2.0.1`, frozen | generic message | `modifiers: :root > #abbrev "2.0.0" -> "2.0.1"` |
| `registries.npm` gains credentials, frozen | -- | `registries: npm "https://registry.npmjs.org/" -> "https://***@registry.npmjs.org/"` |

Scratch monorepo (full reify, two workspaces on one catalog entry): the bump re-resolves in one install; an unused entry writes exactly the options line and leaves **both** lockfiles untouched on the next install (vlt.io's platform-gated optional-only diff hides the `saveHidden` branch); frozen passes after each step.

Release note: catalog and other configuration changes are recorded in `vlt-lock.json` by the install that applies them, and `vlt ci` / `--frozen-lockfile` report which option differs and no longer fail on projects that use modifiers.

Out of scope / known residuals:

- **The hidden lockfile's options block stays stale under an optional-only diff.** On vlt.io every install has a 259-node optional-only diff, so `saveHidden` is skipped and `node_modules/.vlt-lock.json` keeps the old options. Nothing reads it except the no-`vlt-lock.json` fallback, which then costs one extra reset.
- **A dev/optional flag drift found in the control**: removing the last prod edge to a node that is still reachable as a dev dependency leaves its lockfile flags at `0` where they were `2`, and a plain `vlt i` does not restore them. Unrelated to catalogs and present on `main`; filed as #1826.

Windows CI (`9dae51ac`): the catalog test is the first in `install.ts` to drive a real node removal through reify, and the remover's `confirm()` hands the moved-aside store dir to a detached process that raced tap's fixture teardown (`EBUSY` on `rmdir`). The catalog tests now use a `RollbackRemove` with a no-op `confirm()`, so the moved-aside dir is swept with the testdir; the actual loader skips dot-prefixed store entries, so the following frozen install never sees it.

## Review round (`ef2d5cf5`, `3bfa702e`, `3cf15914`)

Four defects from the `9dae51ac` review, each reproduced first:

1. **An alias could reuse the lock of the package it replaces** (`ef2d5cf5`). `lockedFits()`'s loose fallback accepted a locked node whose name matched the *edge* name, so after `vlt i abbrev@2.0.0`, `vlt i abbrev@npm:is-number@^2 --lockfile-only` exited 0 with package.json reading `npm:is-number@^2` and the lockfile still on `~npm~abbrev@2.0.0` (zero manifest fetches). It now compares the name the final spec resolves, which is already the jsr registry form for `jsr:` specs, so the existing JSR-alias reuse test is untouched. Regression covers hydrated and bare locked nodes; after the fix the same commands write `prod npm:is-number@^2 ~npm~is-number@2.1.0`.

2. **A cycle member's modifier left no trace in the cycle's identity** (`3bfa702e`). `baseDepID()` drops the modifier as well as the peer suffix, and intra-SCC targets serialize positionally, so `alias1 -> a1 -> b(mod1) -> a1` and `alias2 -> a2 -> b(mod2) -> a2` hashed byte-equal: a1/a2 merged, both importers ended on `b(mod1)` and gc deleted `b(mod2)`. Colors, class keys and the unit now key on the id minus *only* the peer component.

3. **A collapsed cycle renamed itself on the next pass** (`3bfa702e`). A one-class SCC hashed `identity + newline + env`; once its members merged, the surviving singleton self-loop hashed `env` alone. Confirmed on the foo-twins fixture: `peer.839fc16e63a52b8e` then `peer.9b8df9af16d7e7b4` with no dependency change, propagating into parent hashes. Both paths now build the same string. As the review noted, the existing twins test compared the mutated node with itself; it snapshots the id string before the second pass now.

4. **The frozen exemption matched a modifier by name only** (`3cf15914`). `targets(depName)` checked the last breadcrumb name anywhere, so a modifier for `:root > #unused > #abbrev` exempted the root's unrelated direct `abbrev` edge: with `abbrev: 2.0.0` locked, editing package.json to `^2.0.0` left `vlt i --frozen-lockfile --lockfile-only` at exit 0 on a lockfile still reading `prod 2.0.0`, and `removeSatisfiedSpecs()` would not heal it either. `targetsImporterEdge(importer, name)` takes the importer and only counts a breadcrumb whose whole scope is that edge -- a lone `#name`, or an importer selector the node matches; `tryImporter()` shares the selector match. After the fix that frozen run exits 1 naming the spec change and a plain install heals the edge to `^2.0.0`, while `:root > #abbrev` keeps its exemption.

Validation: graph 2416/2416 asserts at 100% coverage (10932 statements, 2244 branches, 274 functions), dep-id 425/425, security-archive 92/92, query `scanned` 2/2, format, lint and `tsc --noEmit` clean. Every regression is red on `9dae51ac`.

On vlt.io (`spike/rec5-settled`, `--lockfile-only`): the first install renames all 282 peer nodes -- the singleton hash now carries the identity prefix -- with 0 real adds or removes, and a signature check over all 1791 nodes (base id + sorted out-edges by base target) reports 0 differences, so the graph is identical modulo the suffix strings. The second install is byte-identical, `--frozen-lockfile` exits 0, a zero-dep add writes exactly one edge line with 0 renames, and the same add with `node_modules` moved aside is byte-identical to the `node_modules`-backed one. The 12 `renamesWithEnvChange` the differ reports on the migration install are its heuristic same-base pairing: they come in swapped pairs (two `vite` copies, three `use-sync-external-store` copies), and the signature check shows no environment moved.

Still open, and not claimed otherwise: **canonicalization is representation-consistent for these cases, not a proven fixpoint**. The SCC path now uses one identity and one serialization, and the reproduced oscillations are gone, but there is no general order-independence proof; the peer-context limitations below stay follow-ups.

Release note: peer dependency copies get content-hashed DepIDs, so **the first install after upgrading rewrites the peer entries of an existing `vlt-lock.json` once and moves the matching `node_modules/.vlt` store directories**. Verified on vlt.io: `--frozen-lockfile` still exits 0 -- the check compares specs and options, not peer ids -- but it rewrites the file too, so run a plain `vlt install` and commit the result rather than letting CI produce it. No package version or edge target changes; later installs are no-ops.

## Summary by CodeRabbit

* **Improvements**
  * Improved dependency graph rebuilding, peer dependency handling, stable ordering, and lockfile resolution reuse.
  * Improved preservation of package metadata, locked versions, dependency specifications, and peer-store locations.
  * Improved handling of scoped packages, dangling dependencies, and dependency installation rules.
  * Improved security archive support for peer-specific packages and supported registry aliases.

* **Bug Fixes**
  * Fixed lockfile ordering and peer information preservation during save and reload.
  * Improved detection and reporting of configuration and catalog changes, including clearer frozen-lockfile diagnostics.
  * Fixed importer manifests and lockfiles not consistently reflecting updated dependency specifications.


ruyadorno:
@coderabbitai full review

